PR #2, opened 2015-05-04
basic funcionality + ssl

jsvd:
this pr implements logstash-input-http using Puma server.
Has basic functionality of handling HTTP requests and also supports SSL through java keystores.

### DIFF
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -27,7 +27,7 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
   config :port, :validate => :number, :default => 8080
 
   # Maximum number of threads to use
-  config :max_threads, :validate => :number, :default => 4
+  config :threads, :validate => :number, :default => 4
 
   # SSL Configurations
   #
@@ -60,7 +60,7 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
       @server.add_tcp_listener(@host, @port)
     end
     @server.min_threads = 0
-    @server.max_threads = @max_threads
+    @server.max_threads = @threads
   end # def register
 
   def run(queue)
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -52,6 +52,9 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
   def register
     @server = ::Puma::Server.new(nil) # we'll set the rack handler later
     if @ssl
+      if @keystore.nil? || @keystore_password.nil?
+        raise(LogStash::ConfigurationError, "Settings :keystore and :keystore_password are required because :ssl is enabled.")
+      end
       ctx = Puma::MiniSSL::Context.new
       ctx.keystore = @keystore
       ctx.keystore_pass = @keystore_password.value
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -1,0 +1,100 @@
+# encoding: utf-8
+require "logstash/inputs/base"
+require "logstash/namespace"
+require "stud/interval"
+require "socket" # for Socket.gethostname
+require "puma/server"
+require "puma/minissl"
+
+class Puma::Server
+  # ensure this method doesn't mess up our vanilla request
+  def normalize_env(env, client); end
+end
+
+# This output opens a web server and listens to HTTP requests,
+# converting them to LogStash::Event instances.
+#
+class LogStash::Inputs::Http < LogStash::Inputs::Base
+  config_name "http"
+
+  # If undefined, Logstash will complain, even if codec is unused.
+  default :codec, "plain"
+
+  # Which host or ip to bind to
+  config :host, :validate => :string, :default => "0.0.0.0"
+
+  # Which TCP port to bind to
+  config :port, :validate => :number, :default => 8080
+
+  # Maximum number of threads to use
+  config :max_threads, :validate => :number, :default => 4
+
+  # SSL Configurations
+  #
+  # Enable SSL
+  config :ssl, :validate => :boolean, :default => false
+
+  # The JKS keystore to validate the client's certificates
+  config :keystore, :validate => :path
+
+  #TODO: config :cacert, :validate => :path
+
+  # Set the truststore password
+  config :keystore_password, :validate => :password
+
+  # useless headers puma adds to the requests
+  # mostly due to rack compliance
+  REJECTED_HEADERS = ["puma.socket", "rack.hijack?", "rack.hijack", "rack.url_scheme", "rack.after_reply", "rack.version", "rack.errors", "rack.multithread", "rack.multiprocess", "rack.run_once", "SCRIPT_NAME", "QUERY_STRING", "SERVER_PROTOCOL", "SERVER_SOFTWARE", "GATEWAY_INTERFACE"]
+
+  RESPONSE_HEADERS = {'Content-Type' => 'text/html'}
+
+  public
+  def register
+    @server = ::Puma::Server.new(nil) # we'll set the rack handler later
+    if @ssl
+      ctx = Puma::MiniSSL::Context.new
+      ctx.keystore = @keystore
+      ctx.keystore_pass = @keystore_password.value
+      @server.add_ssl_listener @host, @port, ctx
+    else
+      @server.add_tcp_listener @host, @port
+    end
+    @server.min_threads = 0
+    @server.max_threads = @max_threads
+  end # def register
+
+  def run(queue)
+    @server.app = Proc.new do |req|
+      begin
+        REJECTED_HEADERS.each {|k| req.delete(k) }
+        req = lowercase_keys(req)
+        body = req.delete("rack.input")
+        @codec.decode(body.read) do |event|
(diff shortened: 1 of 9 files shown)


lib/logstash/inputs/http.rb
```diff
@@ -0,0 +1,100 @@
...
+  default :codec, "plain"
+
+  # Which host or ip to bind to
+  config :host, :validate => :string, :default => "0.0.0.0"
+
+  # Which TCP port to bind to
+  config :port, :validate => :number, :default => 8080
+
+  # Maximum number of threads to use
+  config :max_threads, :validate => :number, :default => 4
```
ph: Even if this class doesn't inherits from `LogStash::Inputs::Threadable` we should use the same option `threads`

ph: We are also using the `workers` concept for the filter/outputs.

jsvd: but since this is a Puma-specific setting, maybe I should change this to a more puma-specific name like puma_max_threads?

lib/logstash/inputs/http.rb
```diff
@@ -0,0 +1,100 @@
...
+  end # def run
+
+  private
+  def lowercase_keys(hash)
+    new_hash = {}
+    hash.each_pair do |k,v|
+      new_hash[k.downcase] = v
+    end
+    new_hash
+  end
```
ph: You can use `#each_with_object` If you need to save your data in a new hash.

```ruby
m = { "awe" => 1, "weeAAAAee" => 2 }
m.each_with_object({}) { |(k,v), memo| memo[k.downcase] = v }
#=> {"awe"=>1, "weeaaaaee"=>2}

File: lib/logstash/inputs/http.rb [hunk 3]
```diff
@@ -0,0 +1,100 @@
...
+      @server.add_tcp_listener @host, @port
+    end
+    @server.min_threads = 0
+    @server.max_threads = @max_threads
+  end # def register
+
+  def run(queue)
+    @server.app = Proc.new do |req|
+      begin
+        REJECTED_HEADERS.each {|k| req.delete(k) }
```
ph: I am wondering if it would be better to takes the headers from the req and to check if the `REJECTED_HEADERS` include it. I assume we have usually less keys in the request than the rejected list, I don't know rack enough if it is the case.

```ruby
m.delete_if { |k,v| REJECTED_HEADERS.include?(k) }
```

lib/logstash/inputs/http.rb
```diff
@@ -0,0 +1,100 @@
...
+  def register
+    @server = ::Puma::Server.new(nil) # we'll set the rack handler later
+    if @ssl
+      ctx = Puma::MiniSSL::Context.new
+      ctx.keystore = @keystore
+      ctx.keystore_pass = @keystore_password.value
+      @server.add_ssl_listener @host, @port, ctx
+    else
+      @server.add_tcp_listener @host, @port
+    end
```
ph: Missing the ()
```ruby
@server.add_tcp_listener(@host, @port)
```

lib/logstash/inputs/http.rb
```diff
@@ -0,0 +1,100 @@
...
+  #TODO: config :cacert, :validate => :path
+
+  # Set the truststore password
+  config :keystore_password, :validate => :password
+
+  # useless headers puma adds to the requests
+  # mostly due to rack compliance
+  REJECTED_HEADERS = ["puma.socket", "rack.hijack?", "rack.hijack", "rack.url_scheme", "rack.after_reply", "rack.version", "rack.errors", "rack.multithread", "rack.multiprocess", "rack.run_once", "SCRIPT_NAME", "QUERY_STRING", "SERVER_PROTOCOL", "SERVER_SOFTWARE", "GATEWAY_INTERFACE"]
+
+  RESPONSE_HEADERS = {'Content-Type' => 'text/html'}
```
ph: Since we aren't returning html, we should use `text/plain`

lib/logstash/inputs/http.rb
jsvd: well. complexity wise it's the same, you have to do MxN checks

lib/logstash/inputs/http.rb
jsvd: :+1:

lib/logstash/inputs/http.rb
```diff
@@ -0,0 +1,100 @@
...
+        @codec.decode(body.read) do |event|
+          event["headers"] = req
+          decorate(event)
+          queue << event
+        end
+        ['200', RESPONSE_HEADERS, ['ok']]
+      rescue => e
+        @logger.error("unable to process event #{req.inspect}. exception => #{e.inspect}")
+        ['500', RESPONSE_HEADERS, ['internal error']]
+      end
```
ph: Thinking out loud here.

Since our http input is in reality a small webservice, we should consider the response format.

If we return a response to the callback we should use the same format as the request.
If they post a json document we should return a json document, same situation with xml.

To make things less complicated.. I dont think we need to return a string, the http code are enough in this case and more flexible.

lib/logstash/inputs/http.rb
ph: Even if it is puma specific, for a user more threads (or more workers) mean more throughput. They don't have to worry about the implementation details this is why I don't think we should use `puma_*`.

lib/logstash/inputs/http.rb
jsvd: yep this is something I want do take care of next: properly use the request content-type to choose the codec and adapt the response.

lib/logstash/inputs/http.rb
ph: Also for the number of threads, what is a good number? Should we make that assumption with the number of CPUs?

Concurrent-ruby provides a cross platform way of getting that information. https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/utility/processor_count.rb

lib/logstash/inputs/http.rb
```diff
@@ -0,0 +1,100 @@
...
+
+  public
+  def register
+    @server = ::Puma::Server.new(nil) # we'll set the rack handler later
+    if @ssl
+      ctx = Puma::MiniSSL::Context.new
+      ctx.keystore = @keystore
+      ctx.keystore_pass = @keystore_password.value
+      @server.add_ssl_listener @host, @port, ctx
+    else
```
ph: in the SSL context, `@keystore*` becomes mandatory, we should add some sort of validation.

lib/logstash/inputs/http.rb
ph: ok created a ticket for this #3

lib/logstash/inputs/http.rb
```diff
@@ -0,0 +1,100 @@
...
+  # Enable SSL
+  config :ssl, :validate => :boolean, :default => false
+
+  # The JKS keystore to validate the client's certificates
+  config :keystore, :validate => :path
+
+  #TODO: config :cacert, :validate => :path
+
+  # Set the truststore password
+  config :keystore_password, :validate => :password
```
ph: Concerning keystore/keystore_password can we use the same option name as the lumberjack input?

ph:
Minor comment on the SSL, LGTM, tested the normal http not the ssl version.

lib/logstash/inputs/http.rb
jsvd: it will eventually support simple .crt files as well so keystores are a separate config parameter

lib/logstash/inputs/http.rb
jsvd: I usually avoid inject or each with object as I find them hard to read and reason about. strong feelings on this?

jsvd:
Addressed most of the issues, @ph 
Thanks to the issues you raised, the specs I wrote showed the teardown sequence wasn't properly closing the tcp socket. it's now fixed!

lib/logstash/inputs/http.rb
ph: No strong feeling on this, Since they are party of ruby I've been using them.

lib/logstash/inputs/http.rb
ph: Maybe point to some doc to setup ssl? I've personally never used keystore jks before.

ph:
wohoo :)
LGTM! :shipit:

For a future enhancement:

@jsvd I was tinkering with the input this weekend, I think we should add a way to support tagging of incoming web calls. I see something like this on the caller side.

```
curl 127.0.0.1:8080/?tags=github,logstash
#OR
curl 127.0.0.1:8080/tags/github,logstash
```

And we add the tags to the generated events so we use 1 input for multiple incoming webhooks, I think this could be a future enhancement.

jsvd:
@ph yep tagging is essential for conditionals. One very nice way of tagging would be to use the headers on the caller side. Since headers will be exploded to the "headers" field they can then be used in conditionals automatically.
Alternatively, we can have a config setting that, when turned on, will parse the query_parameters and construct an hash out of them.

ph:
@jsvd headers work if the user control the caller part, yes we need them +1 :)
@logstash-plugins/owners I think we should include this plugin as part of the defaults plugins.

Issue #3, opened 2015-05-08
Make sure we correctly handle response type


ph:
The current response is always of type text/plain, we should return the same type of the request or follow the content of the `http/accept` header.

jsvd:
implemented by #4

PR #4, opened 2015-05-15
allow applying codecs depending on content-type

jsvd:
implements #3

### DIFF
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -42,6 +42,10 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
   # Set the truststore password
   config :keystore_password, :validate => :password
 
+  # here you can describe how to decode specific content-types
+  # by default the default codec will be used
+  config :additional_codecs, :validate => :hash, :default => { "application/json" => "json" }
+
   # useless headers puma adds to the requests
   # mostly due to rack compliance
   REJECTED_HEADERS = ["puma.socket", "rack.hijack?", "rack.hijack", "rack.url_scheme", "rack.after_reply", "rack.version", "rack.errors", "rack.multithread", "rack.multiprocess", "rack.run_once", "SCRIPT_NAME", "QUERY_STRING", "SERVER_PROTOCOL", "SERVER_SOFTWARE", "GATEWAY_INTERFACE"]
@@ -64,6 +68,10 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
     end
     @server.min_threads = 0
     @server.max_threads = @threads
+    @codecs = Hash.new(@codec)
+    @additional_codecs.each do |content_type, codec|
+      @codecs[content_type] = LogStash::Plugin.lookup("codec", "json").new
+    end
   end # def register
 
   def run(queue)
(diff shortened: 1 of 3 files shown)


lib/logstash/inputs/http.rb
```diff
@@ -72,7 +80,7 @@ def run(queue)
         REJECTED_HEADERS.each {|k| req.delete(k) }
         req = lowercase_keys(req)
         body = req.delete("rack.input")
-        @codec.decode(body.read) do |event|
+        @codecs[req["content_type"]].decode(body.read) do |event|
```
ph: I think HTTP send a `content_type` everytime.. But I am not sure at 100%. 
Can this value be nil? Should we use `.fetch("content_type", "text/plain")` to guard agains't this possibility?

ph: Also should we limits the available codecs or we keep the door open to use any codecs? Can it be exploited with malicious code?

ph: Nevermind last comment! I've misread the register.

spec/inputs/http_spec.rb
```diff
@@ -12,6 +13,32 @@
...
+        Thread.new { subject.run(queue) }
+        agent.post!("http://localhost:8080/meh.json",
+                    :headers => { "content-type" => "application/json" },
+                    :body => { "message_body" => "Hello" }.to_json)
+        event = queue.pop
+        expect(event["message_body"]).to eq("Hello")
+      end
+    end
+  end
+
```
ph: Maybe, adding an example to explicitly test the user defined codec?

lib/logstash/inputs/http.rb
jsvd: if the content_type is nil then @codecs will return the text/plain codec due to
`@codecs = Hash.new(@codec)`

ph: Yeah the default values.

spec/inputs/http_spec.rb
ph: A test with and non existing codec too would be great, I guess it will throw an exception but we could probably give the user a friendlier message?

ph:
Current test run, added a few suggestions for test cases.

jsvd:
Issues addressed, thanks for the comments @ph  there was an actual bug found by the new spec :dancer:

ph:
LGTM, ship it!

I have also tried with a non existing codec and the message is fine for the user I think:

```
   Failure/Error: subject.register
     LogStash::PluginLoadingError:
       Couldn't find any codec plugin named 'moredec'. Are you sure this is correct? Trying to load the moredec codec plugin resulted in this error: no such file to load -- logstash/codecs/moredec
```

lib/logstash/inputs/http.rb
```diff
@@ -42,6 +42,10 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
   # Set the truststore password
   config :keystore_password, :validate => :password
 
+  # here you can describe how to decode specific content-types
+  # by default the default codec will be used
+  config :additional_codecs, :validate => :hash, :default => { "application/json" => "json" }
```
jordansissel: I like the feature and the idea, but one concern is that codecs can be configured, and this only lets you refer to codecs by name, not by configuration. For example, `line { charset => "ISO8859-1" }` can't be done with this implementation.

Thoughts?

jsvd: I think this can evolve in a bwc way into 

simple case (as-is): `{ "application/json" => "json" }`

or more complex:  `{ "application/json" => { "json" => { codec_option => "meh" } }`

PR #7, opened 2015-06-16
Basic auth support

jsvd:


lib/logstash/inputs/http.rb
```diff
@@ -75,11 +83,20 @@ def register
   end # def register
 
   def run(queue)
```
ph: Is there any reason why we don't use rack middleware for this kind of authorization?

We are already using the rack api for all the answers I think it would be more flexible to rely on rack for this kind of behaviors. See https://github.com/rack/rack/blob/master/lib/rack/auth/basic.rb, The library is well tested and maintained.

Also we could probably support multipart post with it?

What do you think?

ph: Also as we move to an API based backend we will probably use some sort of rack compatible controllers like sinatra.

ph: I think we should log unauthorized access

ph: Another point here, _Basic Auth_ is the MVP we want but in a future revision we could also support oAuth style tokens,  and rack libraries exist for that https://github.com/nov/rack-oauth2

jsvd: :+1: on logging unauthorized access. also +1 on many other things we can add.
Even with basic auth it would be nice to add different user/password to different "realms" or fractions of the url namespace.

This could be done with an hash
```
{ 
  regexp1 => [user1, pass1],
  regexp2 => [user2, pass2],
  ...
}
```

jsvd: regarding the rack vs custom code, I just made the decision since rack forces a lot of decisions on you and seemed overkill for this, but I can take a second look.

ph: You can provide a block to the Auth::Basic to support multiple credentials.
```ruby
use Rack::Auth::Basic, "secret" do |username, password|
  username == 'foo' && password == 'bar'
end
```

I don't think its overkill, it gives us flexibility and rack we will get security upgrades when they come.

ph: I think you can mount different rack handler with different endpoints.. I believe Rails support it that way.

jsvd: done, thought I spent more time than I'm willing to admit fighting scopes..

ph: Scope is the root of all evil.

lib/logstash/inputs/http.rb
```diff
@@ -91,8 +102,17 @@ def run(queue)
         ['500', RESPONSE_HEADERS, ['internal error']]
       end
     end
+
+    auth = Proc.new do |username, password|
+      username == @user && password == @password.value
+    end if (@user && @password)
```
ph: Personal preference
```
if `basic_auth?`
```

ph:
@jsvd OK, I've looked at the code, I am fine with the rack implementation. The tests are running!
LGTM. @jsvd now I can remove my `.htpasswd` from my nginx setup.. ;)

PR #8, opened 2015-06-28
Improve docs and asciidoc formatting

suyograo:


lib/logstash/inputs/http.rb
```diff
@@ -11,26 +11,44 @@ class Puma::Server
   def normalize_env(env, client); end
 end
 
-# This output opens a web server and listens to HTTP requests,
-# converting them to LogStash::Event instances.
+# Using this input you can receive single or multiline events over http(s).
+# Applications can send a HTTP POST request with a body to the endpoint started by this
+# input and Logstash will convert them into events for subsequent processing. Users 
+# can pass plain text, JSON, or any formatted data and use a corresponding codec with this
+# input. By default, the codec used is plain.
```
jsvd: plain for "text/plain" and json for "application/json"

lib/logstash/inputs/http.rb
```diff
@@ -11,26 +11,44 @@ class Puma::Server
...
+# This input can also be used to receive webhook requests to integrate with other services
+# and applications. By taking advantage of the vast plugin ecosystem available in Logstash
+# you can trigger actionable events right from your application.
+# 
+# ==== Security
+# This plugin supports standard HTTP basic authentication headers to identify the requester.
+# You can pass in an username, password combination while sending data to this input
+#
+# You can also setup SSL and send data securely over https, with an option of validating 
+# the client's certificate.
```
jsvd: It might be work mentioning that for now the certificate setup is done using Java Keystores (.jks files) and possible an external guide to generate a jks store?

lib/logstash/inputs/http.rb
jsvd: actually, json for "application/json" and "text/plain" for anything else

lib/logstash/inputs/http.rb
```diff
@@ -11,26 +11,44 @@ class Puma::Server
   def normalize_env(env, client); end
 end
 
-# This output opens a web server and listens to HTTP requests,
-# converting them to LogStash::Event instances.
+# Using this input you can receive single or multiline events over http(s).
+# Applications can send a HTTP POST request with a body to the endpoint started by this
+# input and Logstash will convert them into events for subsequent processing. Users
```
jsvd: "will convert it into an event"

lib/logstash/inputs/http.rb
```diff
@@ -11,26 +11,44 @@ class Puma::Server
   def normalize_env(env, client); end
 end
 
-# This output opens a web server and listens to HTTP requests,
-# converting them to LogStash::Event instances.
+# Using this input you can receive single or multiline events over http(s).
```
jsvd: Not sure what is a multiline event in this context, a plaintext body with some "\n" in the middle?

suyograo: right..plain text with a delimiter that can be chunked into separate events

jsvd:
LGTM

PR #15, opened 2015-07-21
Peg version to 2.11 to bring back rack dependency

suyograo:
This plugin has a dependency on rack framework and is also failing numerous tests. https://github.com/jsvd/logstash-input-http/commit/f433f68539e3836ac99f432ba6718e4a231d6008


jordansissel:
Can you update the version-change commit message to include the reason for the version pinning?

suyograo:
@jordansissel done, when squashing I will rewrite the message

jordansissel:
Otherwise code lgtm! Haven't tested.

Issue #17, opened 2015-07-23
uninitialized constant LogStash::Inputs::Http::Base64


arabold:
I'm using Logstash 1.5.3 and was trying a very simple Logstash configuration. However, when starting up Logstash throws an error: `uninitialized constant LogStash::Inputs::Http::Base64`.

Here's my config:
```
input {

  syslog {
    type => "syslog"
    port => 5544
    tags => []
    add_field => {
      "index" => "syslog"
    }
  }

  http {
    type => "http"
    port => 8080
    user => "logstash"
    password => "*******"
    tags => [ "webhook" ]
  }

}

# Add your filters here
filter {

}

output {
  # Trying to debug things? Enable the stdout output plugin
  stdout {
    codec => rubydebug
  }
}
```

What am I doing wrong here?

suyograo:
Fixed via https://github.com/logstash-plugins/logstash-input-http/pull/18. Thanks @wiibaa

PR #18, opened 2015-07-24
Missing require for base64

wiibaa:
As reported in #17, http basic auth is unusable due to this missing required.
This is not catched by the specs because base64 is transitively loaded when requiring ftw (used as a dev dependency)


suyograo:
LGTM

PR #30, opened 2015-08-18
Add port randomization to specs

### LINKED ISSUE #28: Tests do no port randomization
http://build-eu-00.elastic.co/view/LS%20Inputs%20New/job/logstash-plugin-input-http-unit/5/jdk=JDK8,nodes=metal-pool/console

jsvd:
uses rspec random seed to randomize the tcp ports used in each test

fixes #28

### DIFF
--- a/spec/inputs/http_spec.rb
+++ b/spec/inputs/http_spec.rb
@@ -6,20 +6,25 @@ require "stud/temporary"
 
 describe LogStash::Inputs::Http do
 
+  before do
+    srand(RSpec.configuration.seed)
+  end
+
   let(:agent) { FTW::Agent.new }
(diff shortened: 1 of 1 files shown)


spec/inputs/http_spec.rb
```diff
@@ -6,20 +6,25 @@
 
 describe LogStash::Inputs::Http do
 
+  before do
```
purbon: not related to this specs, but I guess on the feature we should make this standard in our helpers, I've used this same pattern so often for now.

jsvd: agreed, this could go into "logstash/devutils/rspec/spec_helper"

purbon:
LGTM

PR #31, opened 2015-08-31
use plugin.stop to return from run

### LINKED ISSUE #32: refactor shutdown sequence to use plugin.stop
Change shutdown sequence to be triggered by plugin.stop instead of ShutdownSignal exception.
Also remove any calls to: shutdown, finished, finished?, running? or terminating?
This depends on https://github.com/elastic/logstash/pull/3895

jsvd:
Implements `stop` to return from `run` according to https://github.com/elastic/logstash/issues/3210

Depends on https://github.com/elastic/logstash-devutils/pull/32 and https://github.com/elastic/logstash/pull/3895

resolves #32

### DIFF
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -144,14 +144,12 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
   end
 
   public
-  def teardown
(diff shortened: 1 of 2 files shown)


purbon:
LGTM

PR #37, opened 2015-11-29
add verify_mode option to verify client certs

joemiller:
Adds `verify_mode` option that can be used to authenticate client TLs
certificates.

Options: `peer`, `force_peer`, `none`.

- `peer`: Configures https server to request client certificate, but not
  require it.
- `force_peer`: Require client certificate to be trusted by one of the CA's in
  the keystore.

### DIFF
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -69,6 +69,9 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
   # Set the truststore password
   config :keystore_password, :validate => :password
 
+  # Set the client certificate verification method. Valid methods: none, peer, force_peer
+  config :verify_mode, :validate => ['none', 'peer', 'force_peer'], :default => 'none'
+
   # Apply specific codecs for specific content types.
   # The default codec will be applied only after this list is checked
   # and no codec for the request's content-type is found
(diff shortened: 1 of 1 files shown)


ph:
Hello @joemiller Would you mind rebasing this PR?

I think this will also fixe this issue https://github.com/logstash-plugins/logstash-input-http/issues/46 that might be related to the JRUBY changes that god merged in in 2.16 ?:)

ph:
nevermind since this is a blocker I will rebase it and test it.

joemiller:
I had it on my list to do this weekend or next week since puma accepted the
patch to support certification verification when run in jruby yesterday -
https://github.com/puma/puma/pull/833 - but if you would like to do it
first, that would be awesome!!!

joemiller:
rebased against master, passed unit tests. Note I bumped puma to `3.x` which should include the patch to honor the verify_mode flags.

ph:
@joemiller 2.16 should include the patch too, we are about to ship a web api with the next version of logstash I would prefer if we stick with 2.X. until 3.0 is released for some time ;)


https://github.com/puma/puma/blob/v2.16.0/History.txt#L11

joemiller:
I agree. I did not see that 2.16.0 would also have the patch =) I will
adjust the PR.

joemiller:
@ph adjusted puma dep to 2.16

lib/logstash/inputs/http.rb
```diff
@@ -69,6 +69,9 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
   # Set the truststore password
   config :keystore_password, :validate => :password
 
+  # Set the client certificate verification method. Valid methods: none, peer, force_peer
+  config :verify_mode, :validate => :string, :default => 'none'
```
jsvd: thankfully the config dsl in logstash allows us to specify a short list of possible values, so we can validate the method setting here instead of checking `else` in line 109/110. what do you think?

suggestion:
`config :verify_mode, :validate => ['none', 'peer', 'force_peer'], :default => 'none'`

ph: oops I missed that.

ph:
@jsvd I tested it manually and it fixed the issues people were encountering in #46 

I have also created this https://github.com/logstash-plugins/logstash-input-http/issues/47
So we can have more integration tests concerning SSL scenarios.

lib/logstash/inputs/http.rb
joemiller: Sure, I can make this change.

PR #40, opened 2016-01-08
added support for compressed http requests

dwapstra:
My first ruby code hack... need to add tests still.

I have tested this with curl and it works for me...

I'm unsure about the best place for the 'require' statement, suggestions welcome.

### DIFF
--- a/lib/logstash/util/http_compressed_requests.rb
+++ b/lib/logstash/util/http_compressed_requests.rb
@@ -21,13 +21,26 @@ class CompressedRequests
     end
 
     status, headers, response = @app.call(env)
+    if extracted =~ /^(Gzip|Inflate) decompression failed$/
+      status = 400
+    end
     return [status, headers, response]
   end
   
   def decode(input, content_encoding)
     case content_encoding
-      when 'gzip' then Zlib::GzipReader.new(input).read
-      when 'deflate' then Zlib::Inflate.inflate(input.read)
+      when 'gzip' then
+        begin
+          Zlib::GzipReader.new(input).read
+        rescue Exception
+          "Gzip decompression failed"
+        end
+      when 'deflate' then
+        begin
+          Zlib::Inflate.inflate(input.read)
+        rescue Exception
+          "Inflate decompression failed"
+        end
     end
   end
 end
--- a/lib/logstash/util/http_compressed_requests.rb
+++ b/lib/logstash/util/http_compressed_requests.rb
@@ -1,0 +1,33 @@
+class CompressedRequests
+  def initialize(app)
+    @app = app
+  end
+
+  def method_handled?(env)
+    !!(env['REQUEST_METHOD'] =~ /(POST|PUT)/)
+  end
+
+  def encoding_handled?(env)
+    ['gzip', 'deflate'].include? env['HTTP_CONTENT_ENCODING']
+  end
+  
+  def call(env)
+    if method_handled?(env) && encoding_handled?(env)
+      extracted = decode(env['rack.input'], env['HTTP_CONTENT_ENCODING'])
+
+      env.delete('HTTP_CONTENT_ENCODING')
+      env['CONTENT_LENGTH'] = extracted.bytesize
+      env['rack.input'] = StringIO.new(extracted)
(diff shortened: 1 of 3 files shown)


dwapstra:
The error handling works and response code is 400, however the response body is still "ok". I think the code [here](https://github.com/logstash-plugins/logstash-input-http/blob/master/lib/logstash/inputs/http.rb#L113-L124) needs some updates but I'm not sure how to get the status and change the response to something other than 'ok'.

lib/logstash/util/http_compressed_requests.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+    return [status, headers, response]
+  end
+  
+  def decode(input, content_encoding)
+    case content_encoding
+      when 'gzip' then
+        begin
+          Zlib::GzipReader.new(input).read
+        rescue Exception
```
jsvd: I believe here you can catch Zlib::Error instead of the global exception

lib/logstash/util/http_compressed_requests.rb
```diff
@@ -0,0 +1,46 @@
...
+          "Gzip decompression failed"
+        end
+      when 'deflate' then
+        begin
+          Zlib::Inflate.inflate(input.read)
+        rescue Exception
```
jsvd: same here, catch Zlib::Error instead

dwapstra:
the event is still processed even through there is a decompression error. Not sure if we want to avoid processing the request in this case.

jsvd:
I don't think we should let the event through as it is a malformed request. with that in mind, for the `CompressedRequests` module, I'd suggest something along the lines of:

```ruby
  def call(env)
    if method_handled?(env) && encoding_handled?(env)
      extracted = decode(env['rack.input'], env['HTTP_CONTENT_ENCODING'])

      env.delete('HTTP_CONTENT_ENCODING')
      env['CONTENT_LENGTH'] = extracted.bytesize
      env['rack.input'] = StringIO.new(extracted)
    end

    @app.call(env)

  rescue Zlib::Error
    [400, {'Content-Type' => 'text/plain'}, ["Failed to decompress body"]]
  end

  def decode(input, content_encoding)
    case content_encoding
    when 'gzip' then
      Zlib::GzipReader.new(input).read
    when 'deflate' then
      Zlib::Inflate.inflate(input.read)
    end
  end
```

what do you think?

dwapstra:
I have updated the error handling, I think this is much better. It catches decompression errors and does not pass the request to the event handler

lib/logstash/util/http_compressed_requests.rb
```diff
@@ -0,0 +1,44 @@
...
+        return [400, {'Content-Type' => 'text/plain'}, ["Failed to decompress body"]]
+      end
+
+      env.delete('HTTP_CONTENT_ENCODING')
+      env['CONTENT_LENGTH'] = extracted.bytesize
+      env['rack.input'] = StringIO.new(extracted)
+    end
+
+    status, headers, response = @app.call(env)
+    return [status, headers, response]
```
jsvd: @app.call(env) already returns the array so we can reduce that and use implicit return to change:

```ruby
status, headers, response = @app.call(env)
return [status, headers, response]
```

into just:

```ruby
@app.call(env)
```

lib/logstash/util/http_compressed_requests.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+  
+  def decode(input, content_encoding)
+    begin
```
jsvd: since the rescue block's only job is to re-raise the exception we can remove the whole begin/rescue block to be instead

```ruby
def decode(input, content_encoding)
  case content_encoding
  when 'gzip' then
    Zlib::GzipReader.new(input).read
  when 'deflate' then
    Zlib::Inflate.inflate(input.read)
  end
end
```

jsvd:
@dwapstra great job, I left to minor comments just to make the code cleaner. I'll merge and release the plugin afterwards

dwapstra:
@jsvd thanks! I think I got it now

PR #41, opened 2016-01-13
Change field name.

yfhong:
Field name conflict with collectd json metric.


untergeek:
Changing a field name is a breaking change.  This will need to be mentioned in the Changelog, the version number increased to the next major number, and thoroughly documented.

Can you please explain where the collision is?  I understand that you have two field names colliding, but if they're both string values, why is that a conflict?

PR #52, opened 2016-05-26
Fix rewrite of normalize_env method, only to happen in this plugin and not for all puma usage.

### LINKED ISSUE #51: http plugin blocks stats API
Please post all product and debugging questions on our [forum](https://discuss.elastic.co/c/logstash). Your questions will reach our wider community members there, and if we confirm that there is a bug, then we can open a new issue here.

For all general issues, please provide the following details for fast resolution:

- Version: Logstash 5.0.0 alpha2

If you start logstash with this configuration:

```
input {
    http {
      port => 31311 # default: 8080
      type => "WEBHOOK"
    }
}

filter {
}

output {
  stdout { codec => rubydebug }
}
```

And start logstash:

```
./bin/logstash --http-host=127.0.0.1 --http-port=9601 -f /path/to/logstash.conf
```

The stats API will return:

```
curl http://localhost:9601/_node/stats?pretty
{
  "hostname" : "es-rclarke",
  "version" : {
    "number" : "5.0.0-alpha2"
  }
}
```
rather than the expected jvm / events details.

purbon:
Fixes #51 

@ph your thoughts are very welcome here.

### DIFF
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -8,7 +8,15 @@ require "puma/minissl"
 require "base64"
 require "rack"
 
-class Puma::Server
+
+##
+# We keep the redefined method in a new http server class, this is because
+# in other parts of logstash we might be using puma as webserver, for example
+# in the sinatra part we need thie method to actually return the REQUEST_PATH, 
+# so it can actually infer the right resource to use.
+# Fixes https://github.com/logstash-plugins/logstash-input-http/issues/51
+##
+class HTTPPumaServer < Puma::Server
   # ensure this method doesn't mess up our vanilla request
   def normalize_env(env, client); end
 end
@@ -87,7 +95,7 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
   public
   def register
     require "logstash/util/http_compressed_requests"
-    @server = ::Puma::Server.new(nil) # we'll set the rack handler later
+    @server = ::HTTPPumaServer.new(nil) # we'll set the rack handler later
     if @user && @password then
       token = Base64.strict_encode64("#{@user}:#{@password.value}")
       @auth_token = "Basic #{token}"
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -12,11 +12,11 @@ require "rack"
 ##
 # We keep the redefined method in a new http server class, this is because
 # in other parts of logstash we might be using puma as webserver, for example
-# in the sinatra part we need thie method to actually return the REQUEST_PATH, 
+# in the sinatra part we need this method to actually return the REQUEST_PATH, 
 # so it can actually infer the right resource to use.
 # Fixes https://github.com/logstash-plugins/logstash-input-http/issues/51
 ##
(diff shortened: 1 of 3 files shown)


purbon:
to test this you actually need to use a real logstash as there is where we also use puma.

lib/logstash/inputs/http.rb
```diff
@@ -8,7 +8,15 @@
...
 
-class Puma::Server
+
+##
+# We keep the redefined method in a new http server class, this is because
+# in other parts of logstash we might be using puma as webserver, for example
+# in the sinatra part we need thie method to actually return the REQUEST_PATH, 
+# so it can actually infer the right resource to use.
+# Fixes https://github.com/logstash-plugins/logstash-input-http/issues/51
+##
```
ph: thie -> this

lib/logstash/inputs/http.rb
```diff
@@ -8,7 +8,15 @@
...
+class HTTPPumaServer < Puma::Server
```
suyograo: Maybe HTTPInputPumaServer to be clear?

suyograo: At some point we can consolidate the 2 http servers and we can delegate all requests that come to this input to go to main LS http server

ph:
@purbon I like the way of fixing the issue, cleaner than the monkey patch, LGTM.

lib/logstash/inputs/http.rb
purbon: I agree 100 with this, we shuold consolidate our webserver usage cross the code base. Would be nice to open an issue for this.

purbon:
thanks for your review @ph, not sure why but travis pr test timeout, will investigate this and proceed to merge as this is an important bug.

purbon:
according to my test, the PR always blocks on:

```
skywalker% bundle exec rspec --format documentation
--- jar coordinate com.fasterxml.jackson.core:jackson-annotations already loaded with version 2.7.1 - omit version 2.7.0
--- jar coordinate com.fasterxml.jackson.core:jackson-databind already loaded with version 2.7.1 - omit version 2.7.1-1
Using Accessor#strict_set for specs
Run options: exclude {:redis=>true, :socket=>true, :performance=>true, :couchdb=>true, :elasticsearch=>true, :elasticsearch_secure=>true, :export_cypher=>true, :integration=>true, :windows=>true}

Randomized with seed 50619

LogStash::Inputs::Http
  with :ssl => true
    with :keystore and :keystore_password
      should not raise exception
    without :keystore and :keystore_password
      should raise exception
  with default codec
    when receiving a gzip text/plain request that cannot be decompressed
^C
RSpec is shutting down and will print the summary report... Interrupt again to force quit.
^C%                                                                                                                                                                                 skywalker% 
skywalker% 
```

investigating

purbon:
this issue happens also in master, so looks like a known _not fix jet_ error.

ph:
@purbon Can you create a specific issue for that?

suyograo:
@purbon I kicked off a new build with @jsvd fixes in #54 . all's well. Thanks @jsvd

PR #53, opened 2016-05-27
Update puma dependency to match LS core

suyograo:


purbon:
@andrewvc downgraded core to match with here, I wonder what is the best? be in the kinda old version we use here? or in a newer one? thoughts? @jsvd

jsvd:
puma will now "forever" be a dependency of core, maybe we can remove the puma version restriction or remove the puma dependency completely.

jsvd:
closing as puma is no longer used

PR #54, opened 2016-05-30
refactored tests to ensure puma is up

jsvd:
avoid tests stalling due to plugin not being properly started. sleep until puma says it's ready.
also grouped tests so they're a reuse of the before block


purbon:
LGTM

jsvd:
merged to both master and plugin-api-v1

PR #58, opened 2016-08-01
Add metrics for the http input

purbon:
Add metrics for:
- # events.
- # internal errors

related to https://github.com/elastic/logstash/issues/5607 https://github.com/elastic/logstash/issues/5717

### DIFF
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -94,6 +94,7 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
 
   public
   def register
+    @metric_errors = metric.namespace(:errors)
     require "logstash/util/http_compressed_requests"
     @server = ::HTTPInputWebServer.new(nil) # we'll set the rack handler later
     if @user && @password then
@@ -142,10 +143,12 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
           event.set("headers", req)
           decorate(event)
           queue << event
+          metric.increment(:events)
         end
         ['200', @response_headers, ['ok']]
       rescue => e
         @logger.error("unable to process event #{req.inspect}. exception => #{e.inspect}")
+        @metric_errors.increment(:internal_errors)
         ['500', @response_headers, ['internal error']]
       end
     end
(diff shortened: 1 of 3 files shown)


lib/logstash/inputs/http.rb
```diff
@@ -142,10 +143,12 @@ def run(queue)
           event.set("headers", req)
           decorate(event)
           queue << event
+          metric.increment(:events)
```
ph: I think theses metrics are or should be collected at a higher level?
What is the purpose of this one?

purbon: purpose it to track events_in for the inputs, for such input we could for example have 10 events for 1 http request (as example). are you imagining something namespaced for the input? or something general for all the inputs?

ph: I still think we should remove this; We could wrap the queue sent to a plugin with an instrumented queue that would track theses metrics.

**Rought proposal**

```ruby
class InstrumentedQueue
  def initialize(queue, plugin)
    @queue = queue
    @metric = plugin.metric
  end

  def push(item)
     timer = @metric.timer(:wait_on_queue_ms)
     @queue << item
      timer.stop
     @metric.increment(:events_in)
  end
end
alias_method :<<, :push
```

I believe I have discussed it with @jordansissel

lib/logstash/inputs/http.rb
```diff
@@ -142,10 +143,12 @@ def run(queue)
...
         end
         ['200', @response_headers, ['ok']]
       rescue => e
         @logger.error("unable to process event #{req.inspect}. exception => #{e.inspect}")
+        @metric_errors.increment(:internal_errors)
         ['500', @response_headers, ['internal error']]
```
ph: OK with that after reading the 500 description
```
500 Internal Server Error
A generic error message, given when an unexpected condition was encountered and no more specific message is suitable.

jordansissel:
LGTM

ph:
LGTM

ofeki-neosec:
Hello,
I'd like to add some metrics to this input plugin to cover its current blindspots (mainly the 429 status code whenever the server is busy). 
Is there a reason this ticket was closed?
Should I open another ticket and PR with my suggestion?

Issue #61, opened 2016-09-25
Host field is unconditionally overwritten


magnusbaeck:
It seems the http input unconditionally saves the remote host into the `host` field after the codecs are applied. This is fine with e.g. a plain codec but problematic with e.g. a json codec where the event payload might contain a `host` field itself.

I suggest the http input starts behaving like e.g. the tcp input, which only populates the `host` field if it's unset. Quoting [tcp.rb](https://github.com/logstash-plugins/logstash-input-tcp/blob/master/lib/logstash/inputs/tcp.rb):

    event.set(HOST_FIELD, client_address) unless event.get(HOST_FIELD)

(One could argue that `event.get(HOST_FIELD)` should rather be `event.include?(HOST_FIELD)` to deal with the corner case that the host field is set to a non-truthy value.)

jsvd:
+1, what we can also do is have a `target` property that defaults the metadata setting to the root of the event. Either that or an `override_metadata` that defaults to true

jdmac87:
+1

nlowe:
Is there a workaround for this? I've been experiencing this while trying to setup a centralized logging solution for a few of my applications.

jsvd:
what do y'all think of https://github.com/logstash-plugins/logstash-input-http/pull/68?

this PR keeps the current behaviour but adds two configuration options (with terrible names, I accept suggestions) that allow you to override the default target for both remote host and request headers information.

ebuildy:
let's merge it !!!

PR #68, opened 2017-07-28
allow changing the default target field for host and headers data

jsvd:
fixes https://github.com/logstash-plugins/logstash-input-http/issues/61

for reviewer: I accept suggestions for alternative naming of the two options this PR introduces


jsvd:
merged in fe8b72a

Issue #71, opened 2017-10-11
Customer would like the ability to specify ciphers and protocols allowed


e-mars:
- Version: 5.6
- Operating System: CentOS 6.X

Customer is asking for ability to limit which ciphers can be used and which protocols.
similar to what is available in Filebeat with `cipher_suites` and `supported_protocols` options:
https://www.elastic.co/guide/en/beats/filebeat/current/configuration-output-ssl.html

thanks.

AndrewMcQuerry:
Hi @e-mars,  (I figured I'd jump in here and update this issue directly).

I've done some digging into this plugin to see exactly how the http port is opened/served.  I see that it's using the "puma" library, which I've also dug into.

From what I can tell, the protocols and cipher list are hard-coded within the Puma libraries, as shown below:

Inside the “MiniSSL.java” file, line 161
`String[] protocols =newString[] { "TLSv1", "TLSv1.1", "TLSv1.2"};`

And inside “mini_ssl.c” file, line 178
`SSL_CTX_set_cipher_list(ctx, "HIGH:!aNULL@STRENGTH");`


> Disclaimer: Since I'm not intimately familiar with these libraries, this information could be off-track.   Just hoping to provide some input to anyone who's able to take a quick look at this.


If the above is correct, and the Puma library offers no way to manipulate these values, what would be the alternative to being able to remove TLSv1/TLSv1.1 and certain ciphers from being enabled for this port?

jordansissel:
@andrewvc Puma in JRuby doesn't use any C extensions, so the C code you looked at is not used in Logstash (JRuby).

We can make it work, even if it means patching Puma or changing what provides the http service.

AndrewMcQuerry:
Thanks @jordansissel for clearing that up.

AndrewMcQuerry:
Any idea what the source of the cipher suite is that is enabled for these ports?

AndrewMcQuerry:
Hi all,
Can anyone share what the expected time frame is to decide which solution to go with and then how long it would take to have a patch for us to use?

We're looking into using a Proxy to open up the port facing the LAN, and then having it forward all traffic to the input http plugin assigned directly to a localhost port.  But we see this as a temporary work-around, thus the questions around what we can expect for a fix to make its way into the logstash plugin.


Thanks!

jordansissel:
I have no estimate at this time.

jsvd:
this will arrive with #73

jsvd:
[...]

PR #73, opened 2017-11-02
Move to netty

jsvd:


AndrewMcQuerry:
Hi @jsvd, do you have an estimate for when this is expected to be completed?  Thanks!

jsvd:
@AndrewMcQuerry hey, it shouldn't be much longer, though I'm travelling for work this whole week.

Essentially there's a broken test when processing gziped content in netty for some reason. Also I need to add the documentation for the new options that were introduced. These two should be somewhat quick to solve.

src/main/java/org/logstash/plugins/inputs/http/NettyHttpServer.java
```diff
@@ -0,0 +1,94 @@
...
+            this.namePrefix = namePrefix;
+            group = Thread.currentThread().getThreadGroup();
+        }
+
+        @Override
+        public Thread newThread(Runnable r) {
+            Thread t = new Thread(group, r,
+                    namePrefix + "[T#" + threadNumber.getAndIncrement() + "]",
+                    0);
+            t.setDaemon(true);
```
jsvd: Interesting comment by @original-brownbear in https://github.com/elastic/support-dev-help/issues/2816#issuecomment-347161856:

> @jsvd hmm, (wrong place to discuss this :D), but generally speaking I think it's a big anti-pattern to make I/O threads daemon threads. They pretty much always hold state in their buffers and you want that state to be cleanly resolved.
> If you just make the Daemon threads you have no assurances to that end. If shutdown on interrupting once gets stuck, that means that some thread is simply not joining because the interrupt handling in that thread is wrong.
> Http and beats are a good example here, if you just turn those into daemons, won't you still accept connections from them when the pipeline is already shut down? Seems to me that causes a lot of needless connection reset by peer, when you could just as well kill the server socket first, finish with the client socket handling and then stop the pipeline?

jsvd: > They pretty much always hold state in their buffers and you want that state to be cleanly resolved.

We already control the graceful shutdown sequence by catching sigint and calling the agent.shutdown, which attempts to shutdown everything in a controlled manner, namely doing `processorGroup.shutdownGracefully(0, 10, TimeUnit.SECONDS).sync();` to terminate the nio loop group.

> Http and beats are a good example here, if you just turn those into daemons, won't you still accept connections from them when the pipeline is already shut down?

I'm not sure how is this related to a thread being daemonized or not, this window of accepting connections can be tuned with the grace_period and timeout of that `shutdownGracefully` call

original-brownbear: @jsvd I admit I didn't read the whole code here and was just making general statements and maybe the implementation here takes care of everything, but:

> processorGroup.shutdownGracefully(0, 10, TimeUnit.SECONDS).sync();

who guarantees you that those threads will live for the 10s they need to gracefully shut down if they're all daemon threads and the JVM will stop without joining them?

jsvd: > if they're all daemon threads and the JVM will stop without joining them?

I don't see why the JVM would stop as we're waiting for the closefuture to return by calling sync. That said I'm not familiar enough with the code to understand the shutdown sequence (those eventloop shutdown methods are a rabbit hole).

original-brownbear: @jsvd 

> closefuture to return by calling 

ah that's what I was looking for. Makes sense :) That said, since we're doing that, why do we need to make those thread daemon threads in the first place if we ensure their shutdown anyways?

jsvd: I guess my counter point is the reverse of that question: since we're controlling the shutdown sequence, why allow jvm to block us from terminating (which happened in the `exit(1)` issue with jruby)

original-brownbear: @jsvd 

> why allow jvm to block us from terminating

Fair point, the functional impact here is pretty slim. But imo it's more of a correctness thing:
* If our shutdown handling with the `sync` call and the like works fine => setting `daemon == true` is dead code
* If it's not working properly, we get visibility on that

---------------------

Sorry for wasting time though, I think for HTTP in particular (with its short lived connections) this is very very minor. This is more something to think about in the context of E2E acking/Beats, where it probably makes sense to have that level of formal correctness in this stuff.

jsvd:
pr https://github.com/logstash-plugins/logstash-input-http/pull/75 introduces a 429 response when busy, so this PR must implement that as well before merging.

lib/logstash/inputs/http.rb
```diff
@@ -57,28 +40,61 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
...
-  #
-  # Enable SSL
+  # Events are by default sent in plain text. You can
+  # enable encryption by setting `ssl` to true and configuring
+  # the `ssl_certificate` and `ssl_key` options.
   config :ssl, :validate => :boolean, :default => false
 
-  # The JKS keystore to validate the client's certificates
-  config :keystore, :validate => :path
+  # SSL certificate to use.
```
jakelandis: These SSL settings are bit confusing.  I would suggest to present configuration in the form of keystore and truststore's.  Presumption that both are PKCS#12 backed java keystores, and _may_ be the same file. 

Keystore holds the Pb key (certs) it presents to clients for verification and the Pr key it uses for TLS. 
Truststore holds the Pb keys (certs) for the all the entire chain of trust for mutual (peer) TLS such as the self-signed, intermediate, CA certs, etc. 

specifically : ssl_key -> keystore , ssl_key_passphrase -> keystore_password, ssl_certficate_authorities -> trust_store, (add trust_store_password)

I think we should use this opportunity to default to true for SSL/TLS (and encourage, but not require mutual (peer) TLS).

specifically : ssl -> true (not false) , and ssl_verify_mode -> ssl_peer_verify_mode = [none, verify, force_verify] (with verify the default and warns if client does present a cert)

jakelandis: I see now that these are the same settings as beats, but the recommendation that I made aligns us closer to the ES output (which I think is more correct).  This probably requires more discussion on convergence of configuration options.

src/main/java/org/logstash/plugins/inputs/http/NettyHttpServer.java
```diff
@@ -0,0 +1,74 @@
...
+            daemonThreadFactory("handler-executor"), 15,
+            new RejectedRequestHandler());
+
+    public NettyHttpServer(final String host, final int port, final SslSimpleBuilder sslSimpleBuilder)
+    {
+        this.host = host;
+        this.port = port;
+        this.sslSimpleBuilder = sslSimpleBuilder;
+        serverBootstrap = new ServerBootstrap()
+                .group(processorGroup)
```
jakelandis: Typically you use a boss group and worker group as illustrated in [this](http://netty.io/4.1/xref/io/netty/example/http/snoop/HttpSnoopServer.html#53) example.  If I understand correctly from past discussions, we aren't doing that here to and pushing the worker group to a child handler, to allow that to throw an exception, then handle that exception to set the 429 response code ?

.... and then the SO_BACKLOG is there as a fail safe to reject after too many, such that SO_BACKLOG should always be higher then 429 threshold ?

Just want to understand what is going on here...

src/main/java/org/logstash/plugins/inputs/http/util/DaemonThreadFactory.java
```diff
@@ -0,0 +1,30 @@
...
+
+    DaemonThreadFactory(String namePrefix) {
+        this.namePrefix = namePrefix;
+        group = Thread.currentThread().getThreadGroup();
+    }
+
+    @Override
+    public Thread newThread(Runnable r) {
+        Thread t = new Thread(group, r,
+                namePrefix + "[T#" + threadNumber.getAndIncrement() + "]",
```
jakelandis: can you add "http-input" to the thread name ?

src/main/java/org/logstash/plugins/inputs/http/NettyHttpServer.java
jsvd: > we aren't doing that here to and pushing the worker group to a child handler, to allow that to throw an exception, then handle that exception to set the 429 response code ?

Correct. I played a bit with two groups, and I could not find a way to control the creation of new connections: the first never blocks as its job is just to accept connections and pass them to the second group.
With the single worker we pass the work to the custom executor at the end of the pipeline. If that executor is busy we return 429. If we're too busy returning 429s and the connection requests keep flooding in we have SO_BACKLOG in place to reject connections.

lib/logstash/inputs/http.rb
jsvd: Yeah, I used beats code and settings since it had the most complete set of settings for ssl handling (support for intermediate cas, control over ciphers, tls min/max version setting, etc.

looking at other places in the stack, namely x-pack security, we tend to prefer PEM format over jks.

That said ++ on a broader discussion of common ssl settings

build.gradle
```diff
@@ -0,0 +1,97 @@
...
+    testCompile 'org.hamcrest:hamcrest-library:1.3'
+    testCompile 'org.apache.logging.log4j:log4j-core:2.6.2'
+
+    compile 'io.netty:netty-all:4.1.3.Final'
+    compile 'io.netty:netty-tcnative-boringssl-static:1.1.33.Fork23'
+    compile 'org.javassist:javassist:3.20.0-GA'
+    compile 'com.fasterxml.jackson.core:jackson-core:2.7.5'
+    compile 'com.fasterxml.jackson.core:jackson-annotations:2.7.5'
+    compile 'com.fasterxml.jackson.core:jackson-databind:2.7.5'
+    compile 'com.fasterxml.jackson.module:jackson-module-afterburner:2.7.5'
```
jakelandis: what are the jackson and javaassist dependencies for ?

jsvd: copy paste nonsense.

src/main/java/org/logstash/plugins/inputs/http/NettyHttpServer.java
```diff
@@ -0,0 +1,74 @@
...
+    private final int numProcessors = Runtime.getRuntime().availableProcessors();
+    private final ServerBootstrap serverBootstrap;
+    private final String host;
+    private final int port;
+    private SslSimpleBuilder sslSimpleBuilder;
+
+    private final EventLoopGroup processorGroup =
+            new NioEventLoopGroup(numProcessors, daemonThreadFactory("http-input-processor"));
+
+    private final DefaultEventExecutorGroup executorGroup = new DefaultEventExecutorGroup(4,
```
jakelandis: We probably want to expose the thread and pending tasks as configuration options, and may want to default to some function of the number of cores.

src/main/java/org/logstash/plugins/inputs/http/IMessageHandler.java
```diff
@@ -0,0 +1,25 @@
+package org.logstash.plugins.inputs.http;
+
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.handler.codec.http.FullHttpRequest;
+import io.netty.handler.codec.http.FullHttpResponse;
+
+/**
+ * Created by joaoduarte on 16/10/2017.
```
jakelandis: As an American this date confuses me.  As a programmer, this makes me happy.

src/main/java/org/logstash/plugins/inputs/http/HttpInitializer.java
```diff
@@ -0,0 +1,42 @@
...
+    protected void initChannel(SocketChannel socketChannel) throws Exception {
+        ChannelPipeline pipeline = socketChannel.pipeline();
+
+        if(sslSimpleBuilder != null) {
+            SslHandler sslHandler = sslSimpleBuilder.build(socketChannel.alloc());
+            pipeline.addLast(sslHandler);
+        }
+        pipeline.addLast(new HttpServerCodec());
+        pipeline.addLast(new HttpContentDecompressor());
+        pipeline.addLast(new HttpObjectAggregator(512 * 1024));
```
jakelandis: Does this mean that if someone is trying to stream us chucked data, we have a max chunk size of 512Kb ? Does this 512Kb limit also apply to non-chunked requests ?

Should we add tests for chunked requests?

src/main/java/org/logstash/plugins/inputs/http/HttpServerHandler.java
```diff
@@ -0,0 +1,53 @@
+package org.logstash.plugins.inputs.http;
+
+import io.netty.buffer.ByteBuf;
+import io.netty.channel.*;
+import io.netty.handler.codec.compression.DecompressionException;
+import io.netty.handler.codec.http.*;
```
jakelandis: nitpick: most style guides i have seen recommend against wildcard imports. in IntelliJ just set the min number to 99 in the Java formatter settings.

lib/logstash/inputs/http/message_handler.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    private
+    def generate_response(status, headers, content)
+      payload = Unpooled.copiedBuffer(content.to_java_string, CharsetUtil::UTF_8)
+      response = DefaultFullHttpResponse.new(
+        HttpVersion::HTTP_1_1,
+        HttpResponseStatus.valueOf(status),
+        payload)
+      response.headers().set(HttpHeaderNames::CONTENT_LENGTH, payload.readable_bytes());
+      response.headers().set(HttpHeaderNames::CONTENT_TYPE, "text/plain");
```
jakelandis: May want to consider also supporting json as the content type, flexed by the "Content-Type" header.

src/main/java/org/logstash/plugins/inputs/http/NettyHttpServer.java
```diff
@@ -0,0 +1,83 @@
...
+        this(host, port, messageHandler, sslSimpleBuilder, NUM_PROCESSORS);
+    }
+
+    public NettyHttpServer(String host, int port, IMessageHandler messageHandler,
+                           SslSimpleBuilder sslSimpleBuilder, int threads)
+    {
+        this.host = host;
+        this.port = port;
+        processorGroup = new NioEventLoopGroup(threads, daemonThreadFactory("http-input-processor"));
+        executorGroup = new DefaultEventExecutorGroup(threads, daemonThreadFactory("http-input-handler-executor"),
```
jakelandis: minor: 
it would be swell to include the plugin id in the thread name as well.  If multiple http inputs are present in the same running instance, it is not possible from a stack track/profiler to tell which executors belong to which processors group. 

...that or just move the thread name value incrementing to static scope so that every http-input* has a unique number (and some mental math to associate them).

jakelandis:
@jsvd - I am trying to test a workflow that will respond in a `429` code, but unable to.  Blocking the pipeline with the sleep filter i can effectively block new requests, but rather issuing `429` responses, the clients are simply blocked waiting on a response. New clients can still submit a requests, but those requests also get blocked waiting on a response.

jsvd:
@jakelandis i believe this is due to the executor's maxpendingtasks which defaults to 16. You can use the sleep filter and a curl loop with -m 0.1 to see the 429s

jakelandis:
@jsvd - you are right about the maxpendingtasks...but there is some funkiness that that brings. 

Here are my testing notes:

with `pipeline.workers"=>1, "pipeline.batch.size"=>1`
means that the blocking queue size = 2

with the sleep filter set to a high value
`sleep {  time => "999999"  }`

```
for i in {1..21};do  curl -v  -H "Content-Type: application/json"  -X GET -d '{"abc":"123","dec":456}' http://192.168.1.121:8080 ;done
```

The 4th event is blocked [1] from writing to the queue. (not the 3rd since the 1st is pulled from the queue and is sleeping)

The 4th event is also stuck waiting on a response. If I understand correctly the 4th event shouldn't get stuck waiting for a response, but rather should get a `429` response code. 

What is interesting is that actually does happen, but only AFTER (exactly) 20 requests, where 3 are full request/response, 1 is blocked, and 16 requests timeout without the message ever getting processed.  

```
for i in {1..21};do  curl -v -m 0.1  -H "Content-Type: application/json"  -X GET -d '{"abc":"123"}' http://192.168.1.121:8080 ;done
```
The magic number here is 16, which is indeed `io.netty.eventexecutor.maxPendingTasks` which has min value of 16 (this.maxPendingTasks = Math.max(16, maxPendingTasks);) ...which means that 16 requests from a single client NEED to timeout via the client before back pressure (`429`) starts to kick in. ...or 16 different clients are hung up waiting on a response.... or more acuratley 16 requests have been accepted at the socket layer, but not not processed by the message handler yet. It isn't just the first 16, it is that for backpressure to kick **there will always be need to be 16 unprocessed requests before a `429`'s is returned.** 

If you imagine a single client sending data with retry logic on a read timeout, that client will need to timeout and retry 15 more times before getting `429`'s.  Those 16 will get processed once the pipeline clears up, resulting in duplicates.  

I think if we want `429`'s we need to ensure that is returned for the first blocked request and gauruntee it won't be duplicated if the client re-tries. 


[1]
```
  @@count = 0;
  def decode_body(remote_address, http_full_request, default_codec, additional_codecs)
  ...
      @@count +=  1
      puts "before queue push " + @@count.to_s
      @queue << event
      puts "after queue push"
 ...     
```

jsvd:
> The 4th event is also stuck waiting on a response. If I understand correctly the 4th event shouldn't get stuck waiting for a response, but rather should get a 429 response code.

This is true. It seems the executors naturally set this maxPendingTasks number to > 1 to provide a cushion at the executor's door. Having this set to 1 (or even 16) will likely mean a lot of 429 under high load from multiple sources. We can see this in the current implementation with puma and the blocking queue, with many concurrent requests we see a lot of failures:

numbers/backend | Puma | Netty 
------------ | ---: | ---:
Transactions | 2623501 | 2703234
Transaction rate | 8772.49 | 9036.68 
Failed transactions | 8 | 1
Number of 429 | 1365 | 482

jakelandis:
spoke with @jsvd over slack, and he will try see if it is possible to set the maxPendingTasks to 1 to help ensure that `429` get issued as soon as possible that the pipeline is blocked, and will also help to minimize the client try/timeout/retry duplicate problem.

jsvd:
@jakelandis this should now behave correctly, returning 429 after (queue_size + thread count) requests on a backpressured pipeline. The #queuesize requests will return 200, while all the #threadcount requests will block.

jakelandis:
@jsvd - nice !  did it cost any performance to get rid of that tiny worker queue ? 

@robbavey - FYI https://github.com/logstash-plugins/logstash-input-http/pull/73/commits/bbac6b1d0259aa1a020e6700006a914474d4b762 may also be relevant to the beats input and your POC on implementing queue-less back pressure.  

_I should be able to finish up this review later next week._

jsvd:
> @jsvd - nice ! did it cost any performance to get rid of that tiny worker queue ?

It still uses the ThreadPoolExecutor and uses a SynchronousQueue for the handoff.
Unfortunately this does make thing slower:

SynchronousQueue: 6491.81 trans/sec
ArrayBlockingQueue size 16: 8998.76 trans/sec

AndrewMcQuerry:
Hi all, Any update on when the review will be complete?  Also, how much work would it be for us to test the plugin with the Netty changes so that we can confirm that it covers the functionality we've been waiting for? (custom ciphers and enabled protocols)
Thanks!

jakelandis:
@AndrewMcQuerry - apologies for the delay, Joao is out of office for a bit and this has slipped from my radar.  A new estimate is likely early/mid March for a release.

> custom ciphers and enabled protocols

Is there a specific issue that details the expected changes ? (or is it just for parity with [beats input](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-beats.html) TLS configuration ?)

AndrewMcQuerry:
The previous usage of Puma did not allow us to specific the ciphers and protocols used when enabling SSL on the port for this plugin.   See #71

jsvd:
From what I see the only remaining point to get this pr merged is https://github.com/logstash-plugins/logstash-input-http/pull/73/files#diff-56c1cd86f9e93cecb2dc009c4c7fd815R50

I've spoken with @original-brownbear and @jakelandis and there are conflicting views on which kind of behaviour we should have: using a blocking queue makes the backpressure (429 responses) simpler to observe, while using a sized queue makes the plugin handle bursts better and have less locking.

My options to move forward are:
1) choose either behaviour
2) provide another setting that allows configuring the queue size, where 0 would cause a SynchronousQueue to be used

I know 2) "solves" this argument but I'd like to push for having less options if possible. Thoughts?

jsvd:
I've added a new setting called `queue_size`. it defaults to 0 which creates a SynchronizedQueue, a non 0 value will use a SizedQueue instead. This allows us to test the pros/cons of buffering events

jsvd:
I have also ensured that codecs flush for each request, in the same lines as https://github.com/logstash-plugins/logstash-input-udp/pull/32

@jakelandis I think this is ready for a new review

jakelandis:
@jsvd - what are your thoughts on enabling a TCP keep alive for the batches that are accepted, but not yet processed (like how beats input works) ?

This would help, when a pipeline is blocked, that the client (or a load balancer) doesn't kill the connection and re-send duplicates, but potentially holds open more concurrent connections for longer time periods.

jsvd:
I'm a bit on the fence since the reasons to add this keep alive implementation are:

1. to remedy the [lack of support for timing out writes to the persistent queue](https://github.com/elastic/logstash/issues/8367)
2. to deal with load balancers timing out connections;

The only downside I see of not implementing this is that, in a backpressure situation and where the LB timeout connections, we'll have at most `#threads + queueSize` duplicate requests, which is a small number. 

OTOH, we'll never fully get rid of "duplicates on backpressure+timeout" situation. With keep alives for the active channels we avoid dups when using LB, but any connectivity issue that ends in a timeout will cause duplicates. Even after we have "timeout-able writes" it will still be possible to have duplication, but the window for that to happen will be much smaller.

In terms of performance with a large number of connections, keeping track of this very small and ever changing set of channels with inflight data seems wasteful and potentially expensive, since all channels will have to access the same counter.

jakelandis:
> there are conflicting views on which kind of behaviour we should have

I am flipping my position.  I had originally argued to favor shorter lived connections and rejecting connections with `429` errors ASAP.  I argued that shorter lived connections were kinder to load balancer's and HTTP connection pools (e.g. don't hang all connections from the upstream system).  

However, this behavior of rejecting fast really only works at scale if a) the HTTP clients understand `429` errors b) HTTP client implements increment back offs. If they don't understand a) then the data is likely lost, if they don't understand b) then there is a ton of traffic in a this crazy send/reject loop.  Also it seems there is performance penalty with this approach. 

However, rejections are unavoidable at some point.  I now think that the rejections should happen at some configurable max value to handle spikes, allow for a performative implementation,  and optimize to allow the user avoid the `429` errors at the cost of potential long lived connections. 

I believe that responsibly handling long lived connections should:
* Allow a configurable option for how many connections (or max batches across connections) will hold the connection open before rejection starts. 
* Allow configuration of TCP keep alive to ensure that if the connection is held too long, that it won't idle time out.

jsvd:
> Allow a configurable option for how many connections (or max batches across connections) will hold the connection open before rejection starts.

This in done in this PR through the queueSize, which can be configurable. This queue is global to all connections.

> Allow configuration of TCP keep alive to ensure that if the connection is held too long, that it won't idle time out.

from what I understand we can only [enable or disable keep alive](https://netty.io/4.1/api/io/netty/channel/ChannelOption.html#SO_KEEPALIVE), there's no way to tune other parameters like the time, interval, etc. This PR enables so_keepalive.

jakelandis:
> This in done in this PR through the queueSize, which can be configurable. This queue is global to all connections.

+ 1 , this is was my long winded answer to support keeping the configuration option, but state that supporting a `0` queueSize is probably not needed, and we should default to a reasonable queueSize (maybe 16 * threads ?) 

> from what I understand we can only enable or disable keep alive, there's no way to tune other parameters like the time, interval, etc. This PR enables so_keepalive.

+1 they are configurable via OS settings: http://coryklein.com/tcp/2015/11/25/custom-configuration-of-tcp-socket-keep-alive-timeouts.html , if we wanted more control we _could_ implement the IdleStateHandler like beats input except use empty packets and read vs. write events. 

I think however, we should make this configurable on/off to better align with the MUST statements in the [RFC](https://tools.ietf.org/html/rfc1122#page-101).

jsvd:
> - 1 , this is was my long winded answer to support keeping the configuration option, but state that supporting a 0 queueSize is probably not needed, and we should default to a reasonable queueSize (maybe 16 * threads ?)

ahh I see, even better, I'll remove the ability to make it 0, and default to a larger value 👍

jsvd:
@jakelandis pr updated, I added some more documentation around changing the value of the queue_size. I set the default to 200 since it's what [ES does for their bulk requests queue](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-threadpool.html)

jakelandis:
@jsvd - I submitted a proposal for Netty backed plugin convergence at https://github.com/elastic/logstash/issues/9244

specifically to this PR , I believe the proposed change is

threads -> executor_threads
backlog -> max_concurrent_connections
queue_size -> max_pendings_requests
Java::JavaLang::Runtime.getRuntime.availableProcessors() -> LogStash::Config::CpuCoreStrategy.maximum 

.... but lets discuss over there.

jsvd:
@jakelandis overall I agree with most of those changes, but I'm not sure if the outcome of that decision should be a blocker for this PR to be merged.

that said, I'm ok with doing:

* Java::JavaLang::Runtime.getRuntime.availableProcessors() -> LogStash::Config::CpuCoreStrategy.maximum
* queue_size -> max_pendings_requests

jsvd:
I'm a bit on the fence about `threads -> executor_threads`, for me all threads are "executor threads", we just need to say in the description what they're executing (netty channel pipeline + decoding + write to queue).

I've changed some of the setting names and introduced `max_content_length` (Defaults to 100mb like ES)

@jakelandis More reviews welcome :)

docs/index.asciidoc
```diff
@@ -90,29 +96,32 @@ Apply specific codecs for specific content types.
...
-  * Value type is <<string,string>>
-  * Default value is `"0.0.0.0"`
+  * Value type is <<number,number>>
+  * Default value is `128`
 
-The host or ip to bind
+Sets SO_BACKLOG for the receiving socket, which is the
+maximum queue length for incoming connection requests.
+If a connection request arrives when the queue is full,
+the connection is refused.
```
jakelandis: Do we want to expose this an option ?  If I understand correctly this would only kick in for a HUGE burst of traffic where the payload is queued in the socket. If the pipeline is healthy the payload will get pulled from the socket ASAP and processed, if the pipeline is blocked the payload will get pulled from the socket ASAP and a 429 is returned.

docs/index.asciidoc
```diff
@@ -144,17 +174,97 @@ specify a custom set of response headers
...
+enable encryption by setting `ssl` to true and configuring
+the `ssl_certificate` and `ssl_key` options.
+
+[id="plugins-{type}s-{plugin}-ssl_certificate"]
+===== `ssl_certificate` 
+
+  * Value type is <<path,path>>
+  * There is no default value for this setting.
+
+SSL certificate to use.
```
jakelandis: Can we beef this up a little bit ?

I _think_ this is really a path to a PKCS#8 keystore ?

docs/index.asciidoc
```diff
@@ -144,17 +174,97 @@ specify a custom set of response headers
...
+
+Time in milliseconds for an incomplete ssl handshake to timeout
+
+[id="plugins-{type}s-{plugin}-ssl_key"]
+===== `ssl_key` 
+
+  * Value type is <<path,path>>
+  * There is no default value for this setting.
+
+SSL key to use.
```
jakelandis: Oh ... is the PKCS#8 keystore ?

and `ssl_certificate` is the PEM variant ?  

..I am so confused.

docs/index.asciidoc
```diff
@@ -144,17 +174,97 @@ specify a custom set of response headers
...
+NOTE: This key need to be in the PKCS8 format, you can convert it with https://www.openssl.org/docs/man1.1.0/apps/pkcs8.html[OpenSSL]
+for more information.
+
+[id="plugins-{type}s-{plugin}-ssl_key_passphrase"]
+===== `ssl_key_passphrase` 
+
+  * Value type is <<password,password>>
+  * There is no default value for this setting.
+
+SSL key passphrase to use.
```
jakelandis: > SSL key passphrase to use

SSL key passphrase to use for the `ssl_key`

lib/logstash-input-http_jars.rb
```diff
@@ -0,0 +1,7 @@
+# AUTOGENERATED BY THE GRADLE SCRIPT. DO NOT EDIT.
+
+require 'jar_dependencies'
+require_jar('io.netty', 'netty-all', '4.1.19.Final')
```
jakelandis: Can you roll this back to 4.1.18.Final. 

Beats, TCP and now this will all need to agree upon the version in use. 

https://github.com/logstash-plugins/logstash-input-beats/blob/master/build.gradle#L25

IIRC the mismatch caused an issue with TLS: https://github.com/logstash-plugins/logstash-input-tcp/pull/113 (it may not present here, but best to be safe)

lib/logstash/inputs/http/message_handler.rb
```diff
@@ -0,0 +1,62 @@
...
+
+module LogStash module Inputs class Http
+  class MessageHandler
+    include org.logstash.plugins.inputs.http.IMessageHandler
+
+    attr_reader :input
+
+    def initialize(input, default_codec, additional_codecs)
+      @input = input
+      @default_codec = default_codec
```
jakelandis: I think we need to clone this default_codec

src/main/java/org/logstash/plugins/inputs/http/HttpInitializer.java
```diff
@@ -0,0 +1,46 @@
...
+
+import java.util.concurrent.ThreadPoolExecutor;
+
+/**
+ * Created by joaoduarte on 11/10/2017.
+ */
+public class HttpInitializer extends ChannelInitializer<SocketChannel> {
+    private final IMessageHandler messageHandler;
+    private SslSimpleBuilder sslSimpleBuilder;
+    private int maxContentLength = 100 * 1024 * 1024;
```
jakelandis: can you pass this via the constructor , it looks like you already have it in scope for NettyHttpServer

lib/logstash/inputs/http/message_handler.rb
jsvd: it's done [here](https://github.com/logstash-plugins/logstash-input-http/pull/73/files/117a9e91fe421a9a38d675b5cc7e1bbcecfdac89#diff-67368f447845a8ce55c9184b435bdbc9R47)

src/main/java/org/logstash/plugins/inputs/http/HttpInitializer.java
```diff
@@ -0,0 +1,46 @@
...
+import org.logstash.plugins.inputs.http.util.SslSimpleBuilder;
+
+import java.util.concurrent.ThreadPoolExecutor;
+
+/**
+ * Created by joaoduarte on 11/10/2017.
+ */
+public class HttpInitializer extends ChannelInitializer<SocketChannel> {
+    private final IMessageHandler messageHandler;
+    private SslSimpleBuilder sslSimpleBuilder;
```
jakelandis: may want to pass this via the constructor... something ,something,  immutable

src/main/java/org/logstash/plugins/inputs/http/util/SslSimpleBuilder.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    public File getSslCertificateFile() {
+        return sslCertificateFile;
+    }
+
+    public SslHandler build(ByteBufAllocator bufferAllocator) throws IOException, NoSuchAlgorithmException, CertificateException {
+        SslContextBuilder builder = SslContextBuilder.forServer(sslCertificateFile, sslKeyFile, passPhrase);
+
+        if(logger.isDebugEnabled())
```
jakelandis: were both logging statements intended to be inside the if(logger.isDebugEnabled() ?

parameterized messages are preferable to concats for logging (and often avoids the need for the if(logger.is<>Enabled) check.  For example `logger.debug("Ciphers {}", Arrays.toString(ciphers))`

Also, there is a mix of if`\n`X and if {`\n`X`\n` } , I don't want to start a war, but we should be consistent.

src/main/java/org/logstash/plugins/inputs/http/NettyHttpServer.java
```diff
@@ -0,0 +1,91 @@
...
+
+        if (sslSimpleBuilder != null) {
+            httpInitializer.enableSSL(sslSimpleBuilder);
+        }
+        httpInitializer.setMaxContentLength(maxContentLength);
+
+        serverBootstrap = new ServerBootstrap()
+                .group(processorGroup)
+                .channel(NioServerSocketChannel.class)
+                .childOption(ChannelOption.SO_KEEPALIVE, true)
```
jakelandis: If we enable this, it should be able to be optional turned off.  If we hold the idle connections open (e.g. prevent the client from timing out) we could lock up the upstream system if we are blocked and the upstream system exhausts all of its connections from it's pool.

src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+                            IMessageHandler messageHandler) {
+        this.ctx = ctx;
+        this.req = req;
+        this.remoteAddress = remoteAddress;
+        this.messageHandler = messageHandler;
+    }
+
+    public void onRejection() {
+        /*
+           TODO Should SO_LINGER be used here to force the termination immediately?
```
jakelandis: I don't think so.  HTTP 1.1 should try to reuse the same connection. Also the connection should get closed normally since we are gracefully completing the HTTP request/response life-cycle, and the TIME_WAIT to keep the socket open for business is good thing. 

In Beats input i am pretty sure that SO_LINGER is set to zero since it is TCP only and in times past would forcefully close (RST) the connection.

src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
```diff
@@ -0,0 +1,51 @@
...
+           TODO This should only happen in the event of backpressure which means there's
+           TODO a potential scenario of many connection retries and fd exhaustion
+         */
+        try {
+            final DefaultHttpResponse response = new DefaultHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.TOO_MANY_REQUESTS);
+            response.headers().set(HttpHeaderNames.CONTENT_LENGTH, 0);
+            ctx.writeAndFlush(response);
+        } finally {
```
jakelandis: Can log a throttled message here ?
for example:
```
                long now = System.currentTimeMillis();
                if (now - lastRejectMessage > 60000) {
                    logger.warn("uh oh back pressure! ");
                    lastRejectMessage = now;
                }
```

jakelandis:
@jsvd - sorry for all the noise around SSL settings.  Can we address that in a subsequent PR ? 

Everything else is looking really good, one last round of testing to go and don't want to block this PR on the cosmetics of the SSL settings.

src/main/java/org/logstash/plugins/inputs/http/NettyHttpServer.java
jakelandis: FWIW, it looks like Netty's default (or maybe it my mac ?) is to send keep alives every 60s.  Just a mental note that client side timeout need to be greater then 60s for this to have the desired outcome.

image

jakelandis:
Last round of testing , and all looks good

* backpressure apply based on settings
* recover from backpressure correctly
* correct socket states (no close_wait issues)
* tcp keep alive's sent correctly
* reloads with out thread leaking
* basic auth
* json codec correctly applied when content-type set
* http output to http input with compression and json_batch

I haven't tested the TLS parts yet, but would like to codify those tests after the rest of the TLS settings land. 

I have a few outstanding suggestions for this PR, but no show stoppers.  Feel free to ignore / address at leisure.   LGTM , and will re-review if new commits are added.

jsvd:
@jakelandis addressed most points (outside of ssl remarks), rebased and back to green on ci 👍

src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
jsvd: I'm skeptic in adding this as it can alarm users for something that may happen rather frequently if traffic that is very..bursty. Also, we'd need an equivalent "ok no longer in back pressure", but I don't think it's easy to detect without keeping even more state.

jsvd:
Interesting, I didn't even know .p12 worked :D
I clearly need to look at https://github.com/puma/puma/blob/master/ext/puma_http11/org/jruby/puma/MiniSSL.java and get more functionality into this builder.

Great catch!

lib/logstash-input-http_jars.rb
```diff
@@ -4,4 +4,4 @@
 require_jar('io.netty', 'netty-all', '4.1.18.Final')
 require_jar('io.netty', 'netty-tcnative-boringssl-static', '2.0.7.Final')
 require_jar('org.apache.logging.log4j', 'log4j-api', '2.6.2')
-require_jar('org.logstash.plugins.input.http', 'logstash-input-http', '4.0.0')
+require_jar('org.logstash.plugins.input.http', 'logstash-input-http', '3.1.0')
```
jakelandis: accidental ?

jsvd: it was on purpose, this no longer is a breaking change PR, so I bumped only to 3.1.0 instead of 4.0.0

jakelandis:
.p12 is just my file extension for a pkcs12 keystore, which is the standard java keystore format.  I believe that in the past (current version) it was used for both the trust store (for mutual tls) and the key store (for normal tls).  I am pretty sure the code you just added fixes pkcs12 for the the mutual tls trust (haven't tested yet) , but it is still broken for use as the key store.

jakelandis:
Normal TLS is working as expected, and passive to the prior version. 

However, testing mutual TLS
Good - New config (ssl_verify_mode) with "force_peer" works as expected 
Bad - New config (ssl_verify_mode) with "peer" is working like "force_peer" always requires the certs presented
Bad - Old config (verify_mode) doesn't seem be requiring any certificates, i can present no certs and it still allows. 

```
 http { 
   port => 8080
   ssl => true
   verify_mode => force_peer
   keystore => "/Users/jake/workspace/tls/server/keystore_with_root.p12"
   keystore_password => "foobar"
   #ssl_verify_mode => "force_peer"
   #ssl_key => "/Users/jake/workspace/tls/server/enckey.pkcs8"
   #ssl_key_passphrase => "foobar" 
   #ssl_certificate => "/Users/jake/workspace/tls/server/cert_with_root.pem"
   #ssl_certificate_authorities => "/Users/jake/workspace/tls/ca/cacert.pem"
  }
```

Testing with curl from Linux (curl and ssl and mac haz issues)
(certs are for example.com and point /etc/hosts to my host running LS)
present the certs for mutual auth (force_peer): 
```
curl -v -d "from fedora" https://example.com:8080 --cacert ca/cacert.pem --cert client/cert.pem --key client/key.pem 
```
trust the ca for normal tls (none): 
 ```
curl -v -d "from fedora" https://example.com:8080 --cacert ca/cacert.pem 
```

I did the same tests against the prior version, and it worked as expected.

jsvd:
I've pushed a few changes and I think I solved the not-good situations

> Bad - New config (ssl_verify_mode) with "peer" is working like "force_peer" always requires the certs presented

I used "peer" and was able to get a succesful event through without specifying a client certificate

> Bad - Old config (verify_mode) doesn't seem be requiring any certificates, i can present no certs with force_peer and it still allows.

This is fixed as well, verify_mode was being ignored.

jakelandis:
TLS changes look good ! 

LGTM :shipit:

PR #75, opened 2017-12-08
Prevent file descriptor leaks by limiting the max number of connections with un-returned responses.

andrewvc:
This plugin currently leaks file descriptors. This patch prevents that by changing the behavior such that when all but one webserver threads are occupied a 429 error will be immediately returned. This means that in the event of a blocked queue we will have one thread available to reject all new attempts to write.

As a side benefit, it made sense to fix the concurrency bottleneck around codec so long as we needed the write slots concept by duplicating the codec hash through the write slots

This isn't perfect, we need queue write timeouts to do this well, but it's a decent short term fix.

### DIFF
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -120,16 +120,24 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
       @server.add_tcp_listener(@host, @port)
     end
     @server.min_threads = 0
-    @server.max_threads = @threads
+    # The actual number of threads is one higher to let us reject additional requests
+    @server.max_threads = @threads + 1
     @codecs = Hash.new
+
     @additional_codecs.each do |content_type, codec|
       @codecs[content_type] = LogStash::Plugin.lookup("codec", codec).new
     end
-    @codec_lock = java.util.concurrent.locks.ReentrantLock.new
+
+    @write_slots = java.util.concurrent.ArrayBlockingQueue.new(threads)
+    threads.times do
+      # Freeze these guys just in case, since they aren't threadsafe
+      @write_slots.put(Hash[@codecs.map {|k,v| [k.freeze, v.clone].freeze }.freeze].freeze)
+    end
+
   end # def register
 
+  BUSY_RESPONSE = ['Server busy, please retry request later'.freeze].freeze
   def run(queue)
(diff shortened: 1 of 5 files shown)


lib/logstash/inputs/http.rb
```diff
@@ -138,16 +144,20 @@ def run(queue)
         REJECTED_HEADERS.each {|k| req.delete(k) }
         req = lowercase_keys(req)
         body = req.delete("rack.input")
-        @codec_lock.lock
+        local_codecs = @write_slots.poll()
+        if !local_codecs # No free write slot
+          next [429, {}, ['Write slots full']]
```
robbavey: Any reason why not returning response headers?

Not sure if `write slots full` is a meaningful error message for users?

lib/logstash/inputs/http.rb
```diff
@@ -122,14 +122,20 @@ def register
     @server.min_threads = 0
     @server.max_threads = @threads
     @codecs = Hash.new
+
     @additional_codecs.each do |content_type, codec|
       @codecs[content_type] = LogStash::Plugin.lookup("codec", codec).new
     end
-    @codec_lock = java.util.concurrent.locks.ReentrantLock.new
+
+    num_write_slots = @threads > 2 ? @threads - 1 : 1
```
robbavey: Alternatively, you could do is start the web server with `threads + 1`  to allow `threads` number of write slots, and the extra one for reporting errors, but you've already increased performance...

jordansissel:
Testing: 

```
% seq 1000 | xargs -n1 -P5 -I@ curl -XPOST -D- -H "Content-Type: application/json" localhost:5000/hello/world --data-binary '{ "hello": "world" }'
```

```
% bin/logstash -e 'input { http { port => 5000 } } filter { sleep { time => 1 every => 1 } }'
```

Curl sees this during a transition from happy-to-congested:

```
okHTTP/1.1 200 OK
Content-Type: text/plain
Content-Length: 2

okHTTP/1.1 429 Too Many Requests
Content-Length: 39

Server busy, please retry request laterHTTP/1.1 429 Too Many Requests
Content-Length: 39

Server busy, please retry request laterHTTP/1.1 429 Too Many Requests
Content-Length: 39

```


LGTM

PR #80, opened 2018-03-21
also clone default codec

### LINKED ISSUE #79: Default codec is not thread safe.
When the default codec (`codec =>` ) is used, the decoding is not thread safe. The default codec should be cloned like the `additional_codecs`. 

Work around is to use `additional_codecs` instead of relying on the default `codec =>`.

colinsurprenant:
Fixes #79 

This PR simply support also cloning the default codec. The current specs does test the codec support so I do not believe more specs are required for this specific case which result in a thread safety issue and is hard to test per-se.

### DIFF
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -122,7 +122,9 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
     @server.min_threads = 0
     # The actual number of threads is one higher to let us reject additional requests
     @server.max_threads = @threads + 1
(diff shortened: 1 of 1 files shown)


jsvd:
I see a few failures of builds timing out, but that is happening in master as well.

colinsurprenant:
so yeah, still failing. I am confident these are not related so will go ahead and merge. - I am another Pr incoming for calling flush so I will defer the publishing to the next PR.

colinsurprenant:
this has been released in v3.0.9

PR #81, opened 2018-03-22
flush codec after decode

colinsurprenant:
This PR depends on logstash-plugins/logstash-codec-json_lines#35

The http input only accepts self-contained payload for each request and multipart is not supported. Because of this, we should always flush the codec after each request & codec decode calls to avoid situations where a delimited content missing a final delimiter, for example, would linger in the codec. 

This is similar in nature to the discussion in https://github.com/elastic/logstash/issues/8830#issuecomment-351472669

Spec added to test this condition. Otherwise codec support is also well tested in the current spec.

### DIFF
--- a/spec/inputs/http_spec.rb
+++ b/spec/inputs/http_spec.rb
@@ -136,12 +136,27 @@ describe LogStash::Inputs::Http do
         end
       end
     end
+
     context "with json codec" do
       subject { LogStash::Inputs::Http.new("port" => port, "codec" => "json") }
       it "should parse the json body" do
         agent.post!("http://localhost:#{port}/meh.json", :body => { "message" => "Hello" }.to_json)
         event = queue.pop
         expect(event.get("message")).to eq("Hello")
+      end
+    end
+
+    context "with json_lines codec without final delimiter" do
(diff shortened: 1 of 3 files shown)


spec/inputs/http_spec.rb
```diff
@@ -145,6 +146,20 @@ def do_post
       end
     end
 
+    context "with json_lines codec without final delimiter" do
+      subject { LogStash::Inputs::Http.new("port" => port, "codec" => "json_lines") }
```
jsvd: can't we test this with `line` codec so we don't have to rely on a new feature of json_lines?

colinsurprenant: yeah I guess so.

colinsurprenant:
@jsvd as you suggested I switched to the line codec, good suggestion.

colinsurprenant:
restarted build - seems like the build issues are now solved.

colinsurprenant:
Green. Will squash, bump version, merge and publish.

colinsurprenant:
actually - I will create another PR for the version bump since it will also reference #80.

colinsurprenant:
version bump PR in #82

colinsurprenant:
This has been released in v3.0.9

PR #85, opened 2018-04-30
move most message handling code to java

jsvd:
move the creation of most objects to java land
essentially the ruby side only invokes the codecs

performance bump of 15-20% on a simple noop pipeline ( `bin/logstash -e "input { http {} } output {}"`)

### DIFF
--- a/src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
+++ b/src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
@@ -60,7 +60,7 @@ public class MessageProcessor implements RejectableRunnable {
 
     private FullHttpResponse processMessage() {
         final Map<String, String> formattedHeaders = formatHeaders(req.headers());
-        final String body = readBytes(req.content());
+        final String body = req.content().toString(charset);
         if (messageHandler.onNewMessage(remoteAddress, formattedHeaders, body)) {
             return generateResponse(messageHandler.responseHeaders());
         } else {
@@ -91,13 +91,6 @@ public class MessageProcessor implements RejectableRunnable {
         response.headers().set(HttpHeaderNames.CONTENT_TYPE, "text/plain");
 
         return response;
-    }
-
-
-    private String readBytes(ByteBuf buf) {
-        final byte[] bytes = new byte[buf.readableBytes()];
-        buf.getBytes(0, bytes);
-        return new String(bytes, charset);
     }
 
     private Map<String,String>formatHeaders(HttpHeaders headers) {
--- a/src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
+++ b/src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
@@ -24,7 +24,7 @@ public class MessageProcessor implements RejectableRunnable {
     private final IMessageHandler messageHandler;
     private static final Charset charset = Charset.forName("UTF-8");
 
-    public MessageProcessor(ChannelHandlerContext ctx, FullHttpRequest req, String remoteAddress,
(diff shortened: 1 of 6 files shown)


src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
```diff
@@ -35,11 +44,72 @@ public void onRejection() {
...
+        }
+        response.headers().set(headers);
+        response.headers().set(HttpHeaderNames.CONTENT_LENGTH, payload.readableBytes());
+        response.headers().set(HttpHeaderNames.CONTENT_TYPE, "text/plain");
+
+        return response;
+    }
+
+
+    private String readBytes(ByteBuf buf) {
```
spinscale: just use `ByteBuf.toString(Charset)`

src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
```diff
@@ -35,11 +44,65 @@ public void onRejection() {
...
+            }
             ctx.writeAndFlush(response);
         } finally {
             req.release();
         }
     }
+
+    private FullHttpResponse generateFailedResponse(HttpResponseStatus status) {
+        final FullHttpResponse response = new DefaultFullHttpResponse(req.protocolVersion(), status);
+        response.headers().set(HttpHeaderNames.CONTENT_LENGTH, 0);
```
jakelandis: the prior implementation sent "failed to authenticate" as the response body. 

_fwiw, i don't personally like the plain text response bodies but it could be non-passive to change it_

src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
```diff
@@ -35,11 +44,65 @@ public void onRejection() {
     @Override
     public void run() {
         try {
-            final FullHttpResponse response = messageHandler.onNewMessage(remoteAddress, req);
+            final HttpResponse response;
+            final String token = req.headers().get(HttpHeaderNames.AUTHORIZATION);
+            if (messageHandler.validatesToken(token)) {
```
jakelandis: may want to short circuit this if
```
if (token != null && !token.isEmpty() && messageHandler.validatesToken(token))
```

src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
```diff
@@ -35,11 +44,65 @@ public void onRejection() {
...
+        final FullHttpResponse response = new DefaultFullHttpResponse(
+                req.protocolVersion(),
+                HttpResponseStatus.OK,
+                payload);
+
+        final DefaultHttpHeaders headers = new DefaultHttpHeaders();
+        for(String key : stringHeaders.keySet()) {
+            headers.set(key, stringHeaders.get(key));
+        }
+        response.headers().set(headers);
```
jakelandis: It looks like the authorization header is now being echo'ed back on the response. We should not echo it back since it is just encoded, not encrypted. (`echo amFrZTpmb28= | base64 -D` to see my super secret un/password)

```
 "headers" => {
          "authorization" => "Basic amFrZTpmb28=",
```

src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
jsvd: unfortunately we can't without moving more logic to java, because `validatesToken` has the reponsibility to both validate the token but only if it has been set in the configuration. short circuiting like this would mean that if a user doesn't set up creds in logstash any request without auth will return 401

jakelandis:
LGTM

Issue #35, opened 2015-11-18
Support "Content-Type: application/json; charset=utf-8"


maximpashuk:
I am using logstash 1.5.4.

If I post log message with content-type "application/json" everything works fine.
Hovever, if I use content-type "application/json; charset=utf-8" json message is not processed by logstash (probably because of fallback to plain codec).

Content type with charset specified is standard thing, I think it should be supported by logstash.

My pain is that .NET HttpClient class by default add charset to content-type, so I should explicitly customize HttpClient not to add charset to content-type.

porcobravo:
Faced same issue, but there is an easy workaround to overcome it. In logstash configuration file specify `codec=> json` and it should work smoothly.

hackery:
Encountered the same bug this week - not easy to chase down and broke my QA environment for a day.

https://github.com/logstash-plugins/logstash-input-http/blob/master/lib/logstash/inputs/http.rb#L154

Is this not easily addressed by parsing out the Content-Type into type/subtype and removing any subsequent parameters? Perhaps the correct fallback sequence with regard to the entries in `additional_codecs` should be

* exact match on "type/subtype; parameter[; parameter]*"
* match on "type/subtype"
* default to "plain"

PR #87, opened 2018-05-09
Enforce content-type header parsing

ebuildy:
Use Netty HttpUtil getMimeType method to get the content-type, instead of the raw HTTP header.

This should fix: https://github.com/logstash-plugins/logstash-input-http/issues/35


ebuildy:
maybe I should try first with header, then if no codec is found, try with getMimeType? (for good performance)

jsvd:
merged in 1b76537

PR #93, opened 2018-08-31
avoid creation of SSLContext for each connection

jsvd:
Separate the creation of the SslContext and the SslHandler.
Before, for each new connection an SslContext was created and from it the SslHandler was generated. However creating an SslContext is very expensive due to all the certificate handling.

This PR moves the creation of SslHandler to a SslHandlerProvider that is instantiated with a SslContext instance. This way the SslContext is created once, and from it we get SslHandlers for each new connection.

This largely improves the performance in usecases where ssl is enabled but http keepalive isn't set.


jsvd:
Performance numbers of using http1.1 requests with SSL and `connection=close` (1 connection per request)
logstash 6.4.0 w/ http input 3.0.10 (puma implementation): 103 eps
logstash 6.4.0 w/ http input 3.2.0 (netty implementation):: 299 eps
logstash 6.4.0 w/ this PR: 413 eps

jsvd:
Thanks @danhermann!

> The PR for switching SSL to Netty on the TCP input uses the same approach of creating a single SSLContext rather than one per connection.

Sweet, I should have looked at that for inspiration! Also, I'm certainly taking the opportunity that this is fresh on my mind to go back to reviewing https://github.com/logstash-plugins/logstash-input-tcp/pull/106

PR #95, opened 2018-09-05
fix some verify_mode edge cases

jsvd:
ssl_verify_mode and verify_mode have some bugs:

- setting ssl_verify_mode to none causes NPE
- setting verify_mode to none defaulted to verify_peer
- setting ssl_certificate_authorites and `ssl_verify_mode` to peer/force_peer would still raise exception

The cases are now handled correctly and tested


lib/logstash/inputs/http.rb
```diff
@@ -213,11 +211,16 @@ def build_ssl_params
     return nil unless @ssl
 
     ssl_builder = nil
-    verify_mode_string = nil
+    verify_mode_string = "none"
```
robbavey: Should this be `NONE` - other `verify_mode_string` values are `upcase`d

lib/logstash/inputs/http.rb
```diff
@@ -213,11 +211,16 @@ def build_ssl_params
...
+
+    if original_params.key?("verify_mode")
+      verify_mode_string = @verify_mode.upcase 
+    elsif original_params.key?("ssl_verify_mode")
+      verify_mode_string = @ssl_verify_mode.upcase
+    end
```
robbavey: Does it make sense to put this value into an instance variable to avoid having to check both `ssl_verify_mode` and `verify_mode` in `require_certificate_authorities`?

PR #98, opened 2018-10-23
Cleanup to match best practices in ruby and shell

teancom:
Cleanup both ruby and shell scripts to match various linters'
suggestions. Mostly using the new hash syntax, as well as small
readability changes. Additionally, update gitignore to not accidentally
add in the output from running `rake vendor`

This was done as a "slate-cleaning" and a way to learn the repo before starting any of the issues marked "hacktober".

### DIFF
--- a/lib/logstash/inputs/http.rb
+++ b/lib/logstash/inputs/http.rb
@@ -1,14 +1,13 @@
-# encoding: utf-8
-require "logstash/inputs/base"
-require "logstash/namespace"
-require "stud/interval"
-require "logstash-input-http_jars"
+require 'logstash/inputs/base'
+require 'logstash/namespace'
+require 'stud/interval'
+require 'logstash-input-http_jars'
 
-java_import "io.netty.handler.codec.http.HttpUtil"
+java_import 'io.netty.handler.codec.http.HttpUtil'
 
 # Using this input you can receive single or multiline events over http(s).
 # Applications can send a HTTP POST request with a body to the endpoint started by this
-# input and Logstash will convert it into an event for subsequent processing. Users 
+# input and Logstash will convert it into an event for subsequent processing. Users
 # can pass plain text, JSON, or any formatted data and use a corresponding codec with this
 # input. For Content-Type `application/json` the `json` codec is used, but for all other
 # data formats, `plain` codec is used.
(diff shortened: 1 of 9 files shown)


teancom:
The CI build failed but it appears like it was due to connectivity issues. Is that something that just needs to be kicked off again?

lib/logstash/inputs/http.rb
```diff
@@ -1,74 +1,73 @@
...
   # Codec used to decode the incoming data.
   # This codec will be used as a fall-back if the content-type
-  # is not found in the "additional_codecs" hash
-  default :codec, "plain"
+  # is not found in the 'additional_codecs' hash
+  default :codec, 'plain'
 
   # The host or ip to bind
-  config :host, :validate => :string, :default => "0.0.0.0"
+  config :host, validate: :string, default: '0.0.0.0'
```
jsvd: we tend to stick to => notation for hashes in the logstash and logstash plugins codebase

jsvd:
Thanks for the PR @teancom, I have restarted the tests in travis

lib/logstash/inputs/http.rb
```diff
@@ -243,19 +238,19 @@ def build_ssl_params
   end
 
   def ssl_key_configured?
-    !!(@ssl_certificate && @ssl_key)
+    !(@ssl_certificate && @ssl_key).nil?
```
jsvd: let's skip these changes, IMO it's debatable which one of `!!expr` or `!(expr).nil?` is better

lib/logstash/util/http_compressed_requests.rb
```diff
@@ -4,19 +4,19 @@ def initialize(app)
   end
 
   def method_handled?(env)
-    !!(env['REQUEST_METHOD'] =~ /(POST|PUT)/)
+    !(env['REQUEST_METHOD'] =~ /(POST|PUT)/).nil?
   end
 
   def encoding_handled?(env)
-    ['gzip', 'deflate'].include? env['HTTP_CONTENT_ENCODING']
+    %w[gzip deflate].include? env['HTTP_CONTENT_ENCODING']
```
jsvd: I'd rather stick with an array here, personal preference

Issue #117, opened 2019-12-02
Add option to configure response body


jsvd:
Currently the response body is either the string "ok", or nothing if the `response_code` is set to 204 (no content).

There should be an option to configure a different static response body.

erickjordan:
This would be a great feature to have.

PR #120, opened 2019-12-09
Add reponse_body to be able to provide a custom body (issue 117)

lucabelluccini:
Hello,
I've tried to implement the missing feature as mentioned on https://github.com/logstash-plugins/logstash-input-http/issues/117
I am still not sure how to handle the `content-type` header as it seems we're re-using the one sent in the request, if any.

FYI @jsvd

### DIFF
--- a/src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
+++ b/src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
@@ -23,15 +23,18 @@ public class MessageProcessor implements RejectableRunnable {
     private final String remoteAddress;
     private final IMessageHandler messageHandler;
     private final HttpResponseStatus responseStatus;
+    private final String responseBody;
     private static final Charset charset = Charset.forName("UTF-8");
 
     MessageProcessor(ChannelHandlerContext ctx, FullHttpRequest req, String remoteAddress,
-                            IMessageHandler messageHandler, HttpResponseStatus responseStatus) {
+                            IMessageHandler messageHandler, HttpResponseStatus responseStatus,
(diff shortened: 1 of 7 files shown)


src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
```diff
@@ -88,9 +91,8 @@ private FullHttpResponse generateResponse(Map<String, String> stringHeaders) {
         response.headers().set(headers);
 
         if (responseStatus != HttpResponseStatus.NO_CONTENT) {
-            final ByteBuf payload = Unpooled.wrappedBuffer("ok".getBytes(charset));
+            final ByteBuf payload = Unpooled.wrappedBuffer(this.responseBody.getBytes(charset));
             response.headers().set(HttpHeaderNames.CONTENT_LENGTH, payload.readableBytes());
-            response.headers().set(HttpHeaderNames.CONTENT_TYPE, "text/plain");
```
andsel, line 93: Instead of removing the response's `content-type` we could also provide an customization option for it.

Close to `response_body` it could be also defined a new `response_content_type` which by default contains `text/plain` but in case the user could also modify it. 
In particular on registration phase, if the user customized the response body then it should be also forced to explicitly define also a response's content-type

andsel (changes requested):

src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
lucabelluccini: Yep it's a nice idea. I _hope_ I'll have a chance to resume working on this 😅

PR #122, opened 2020-01-30
experiment in dropping tcnative for jdks ssl

jsvd:
This is an experiment to demonstrate dropping the tcnative dependency for jdk's ssl.

The reason for this removal is that tcnative is sometimes dependent on particular internals of netty, thereby creating (non-explicit) versioning constraints. Also, it is hard to understand the changes that go into tcnative, and coupling that with the removal of CBC ciphers leads to considering not having to rely on this external library for TLS/SSL.


jsvd:
replaced by #126

PR #126, opened 2020-05-12
Update netty/remove tcnative+boringssl dependencies

robbavey:
This commit updates netty dependency to 4.1.49.final, and
additionally removes the dependency on tcnative +
boringssl


src/main/java/org/logstash/plugins/inputs/http/util/SslSimpleBuilder.java
```diff
@@ -116,4 +131,24 @@ private boolean requireClientAuth() {
 
         return false;
     }
+
+    public static String[] getDefaultCiphers(){
+        if (isUnlimitedJCEAvailable()){
+            return DEFAULT_CIPHERS;
+        } else {
+            logger.warn("JCE Unlimited Strength Jurisdiction Policy not installed - max key length is 128 bits");
```
jsvd, line 139: maybe we only need this warning? the one in line 73 seems redundant, or is it a different setting/code path?

robbavey: The intent of the warning on line 73 was to generate log a local warning/error along with the exception being thrown that stops the plugin from starting if no ciphers are available, whereas this is a general warning that lower strength keys will be used

PR #129, opened 2020-10-13
negotiate for credentials when required and not supplied

yaauie:
Before the migration to Netty, the Puma-based backend included the Rack middleware for basic authentication when auth was required. This middleware would intercept requests that did _not_ provide an auth header, and respond with a `401 Unauthorized` that included an `WWW-Authenticate: Basic` header so that the agent (typically a browser) would understand that Basic credentials were required.

This fix restores that behaviour on the new Netty backend.


lib/logstash/inputs/http/message_handler.rb
```diff
@@ -22,6 +22,10 @@ def validates_token(token)
       end
     end
 
+    def requires_token
```
kares, line 25: :open_mouth: never realized JRuby's JI also handles the reverse case (when implementing Java methods in Ruby) for method mapping fromCamel to_underscore

kares (approved):
LGTM, 

would consider having some (`debug`) logging that we're returning an authenticate response.

yaauie:
I've added the requested logging, and will merge when the build is green.

PR #130, opened 2020-10-20
Add example of additional_codecs setting

jaalcorn:
Add an example using the additional_codecs setting to clarify what the document is stating about using two codecs and reinforce the codec setting's tie back to the Content-Type setting


andsel (commented):
Hi @jaalcorn thank's for contributing.
I think the `es_bulk` codec is quite particular, may be worth to see the configuration with a main codec and another more specific, so let understand that the additional_codec are used when no content-type codec is matched, for example:

```ruby
http {
  codec => json
  additional_codecs => {"application/text", "plain"}
}
```

PR #133, opened 2021-01-26
Feat: improved error handling/logging/unwraping

kares:
~ same as https://github.com/logstash-plugins/logstash-input-beats/pull/405 : 
> The unwrapping at the Java level is for [exceptions wrapped by Netty](https://github.com/netty/netty/blob/netty-4.1.49.Final/handler/src/main/java/io/netty/handler/ssl/JdkSslServerContext.java#L288).

> Full exception details will be logged at debug level from the Java side - since we seem to prefer (manual) exception logging at the plugin level. We also make sure to log cause, if any, on the Ruby side which now catches all (expected) Java exceptions.

while also aligns with improvements done [previously add beats'](https://github.com/logstash-plugins/logstash-input-beats/pull/402)

resolves https://github.com/logstash-plugins/logstash-input-http/issues/134


robbavey (approved):
LGTM!

PR #136, opened 2021-05-04
Build: make rake build exit reliably

kares:
reference: https://github.com/elastic/logstash/issues/12869


kares:
merging based on review at https://github.com/logstash-plugins/logstash-input-tcp/pull/175

Issue #123, opened 2020-02-26
Implement ECS-Compatibility Mode


yaauie:
_This is a stub issue, and needs to be fleshed out with details specific to
this plugin._

***

As a part of the effort to make plugins able to run in an ECS-Compatible manner
by default in an upcoming release of Logstash, this plugin needs to either
implement an ECS-Compatibility mode or certify that it does not implicitly use
fields that conflict with ECS.

farrp:
The "host" field should be "host.name". 

The "headers" field in the document root contains many fields that have proper ECS locations:
"headers" => {
"http_version" => "HTTP/1.1",
"request_path" => "/HealthCheck",
"accept_encoding" => "gzip, compressed",
"http_accept" => nil,
"http_user_agent" => "ELB-HealthChecker/2.0",
"http_host" => "172.31.x.x:443",
"request_method" => "GET",
"connection" => "close",
"content_length" => "0"
}

kaisecheng:
| Legacy  | ECS |
| ------------- | ------------- |
| http_version  | http.version  |
| http_user_agent | user_agent.original, user_agent.name, user_agent.version |
| http_host | host.ip, url.port |
| request_method | http.method |
| request_path  | url.path  |
| content_length | http.request.body.bytes |
| content_type | http.request.mime_type |

PR #137, opened 2021-05-28
add ECS support

kaisecheng:
This PR adds ECS support. All legacy `headers` fields map to `[@metadata][http][header]` and some of them  are copied to ECS fields as the following table
| Legacy  | ECS |
| ------------- | ------------- |
| [headers][http_version]  | [http][version] |
| [headers][http_user_agent] | [user_agent][original] |
| [headers][http_host] | [url][domain], [url][port] |
| [headers][request_method] | [http][method] |
| [headers][request_path]  | [url][path] |
| [headers][content_length] | [http][request][body][bytes] |
| [headers][content_type] | [http][request][mime_type] |
| host | [host][ip] |
| headers | [@metadata][http][header] |

Fixed: https://github.com/logstash-plugins/logstash-input-http/issues/123


lib/logstash/inputs/http.rb
```diff
@@ -177,12 +182,35 @@ def decode_body(headers, remote_address, body, default_codec, additional_codecs)
...
+    http_user_agent = headers.get("http_user_agent")
+    http_host = headers.get("http_host")
+    request_method = headers.get("request_method")
+    request_path = headers.get("request_path")
+    content_length = headers.get("content_length")
+    content_type = headers.get("content_type")
+
+    event.set("[http][version]", http_version) if http_version
+    event.set("[user_agent][original]", http_user_agent) if http_user_agent
+    domain, colon, port = http_host.rpartition(":")
```
kares, line 205: not sure this will work as expected e.g.
```
>> 'localhost'.rpartition(":")
=> ["", "", "localhost"]
```
... think we'll have the `host(:port)` part here but feel free to correct me if this returns the raw `Host: host(:port)`,
in either case I think there are cases where it won't work (I am only seeing a test for the `host:port` case maybe having a test for `host` would be useful).

lib/logstash/inputs/http.rb
```diff
@@ -144,6 +146,9 @@ def register
     require "logstash/inputs/http/message_handler"
     message_handler = MessageHandler.new(self, @codec, @codecs, @auth_token)
     @http_server = create_http_server(message_handler)
+
+    @remote_host_target_field ||= ecs_select[disabled: "host", v1: "[host][ip]"]
+    @request_headers_target_field ||= ecs_select[disabled: "headers", v1: "[@metadata][http][header]"]
```
kares, line 151: if this is still a structure containing multiple values I would vote for naming it `headers` :wink: 
in the future (when ECS ships [`http.request.headers`](https://github.com/elastic/ecs/issues/232#issuecomment-776229716)) we will likely expose the field ...

also, for `@metadata` we kind of started using a prefix convention ([example](https://github.com/logstash-plugins/logstash-input-beats/blob/v6.1.3/lib/logstash/inputs/beats.rb#L166)) when names could potentially collide,
e.g. if both http input and filter would be used in the same pipeline and they both exported  `headers` meta-data.
```suggestion
    @request_headers_target_field ||= ecs_select[disabled: "headers", v1: "[@metadata][input][http][headers]"]
```
or rather to make it more explicit, since at some point we might want to expose response headers in meta-data:
```suggestion
    @request_headers_target_field ||= ecs_select[disabled: "headers", v1: "[@metadata][input][http][request][headers]"]
```
wdyt?

spec/inputs/http_spec.rb
```diff
@@ -368,6 +297,113 @@
...
+            expect(event.get("request_headers")).to include("http_host" => "localhost:#{port}")
+            expect(event.get("request_headers")).to include("content_length" => "5")
+            expect(event.get("request_headers")).to include("content_type" => "text/plain")
+          end
+        end
+      end
+    end
+  end
+
+  def setup_client
```
andsel, line 390: The task of `setup_client` is to wait that the http plugin is up and able to respond, so it's not properly a setup of the http client but a wait for server to bootstrap, so rename it with something that expose this behavior is better. Maybe `wait_for_http_is_ready` or something similar

andsel (changes requested):
To me seems ok, I've only asked for minor changes in a couple of points.

kares (changes requested):
:+1: thanks for taking care of this,
suggested a few (naming) changes, let me know what you think.

lib/logstash/inputs/http.rb
kaisecheng: you are right. I thought it returns the default 80 port but it didn't. 
I added a regex to handle `ipv4:port` and `[ipv6]:port`

lib/logstash/inputs/http.rb
kaisecheng: I like the second one to be more explicit in that it is a request header. 👍  thanks for the suggestion

lib/logstash/inputs/http.rb
kaisecheng: Nice catch 🐞

spec/inputs/http_spec.rb
kaisecheng: I rename it to `setup_server_client`. It is doing a setup for the test, so I am not feeling right to say `wait_xx_ready`. Let me know if it doesn't work for you.

lib/logstash/inputs/http.rb
kares: thanks, the new code looks much better.
in terms of the 80 default sorry to cause confusion, I do not think we should set that for 2 reasons:
- if the schema is *https* 80 default would be incorrect
- ecs does not require us to set default values that aren't present

so I think having a `nil` default and not setting `url.port` if it wasn't specified would be better for starters.
at least that's the way LS currently parses URLs into `url.*` with grok patterns.

minor thing to consider would be type-casting port to `Integer` but it's not strictly required, 
however we type-cast `url.port` and others with grok.

kares (approved):
:lollipop: code-wise looking :sunflower:

lib/logstash/inputs/http.rb
```diff
@@ -177,12 +182,50 @@ def decode_body(headers, remote_address, body, default_codec, additional_codecs)
...
+    return if ecs_compatibility == :disabled
+
+    http_version = headers.get("http_version")
+    event.set("[http][version]", http_version) if http_version
+
+    http_user_agent = headers.get("http_user_agent")
+    event.set("[user_agent][original]", http_user_agent) if http_user_agent
+
+    http_host = headers.get("http_host")
+    domain, port = self.class.get_domain_port(http_host)
```
andsel, line 202: I think you could avoid `self.class` if a method is not found in instance, Ruby searches in the class object

andsel (approved):
Just a minor doubt on Ruby syntax

lib/logstash/inputs/http.rb
kaisecheng: `self.class` is required for class method

docs/index.asciidoc
```diff
@@ -32,6 +32,28 @@ This input can also be used to receive webhook requests to integrate with other
 and applications. By taking advantage of the vast plugin ecosystem available in Logstash
 you can trigger actionable events right from your application.
 
+[id="plugins-{type}s-{plugin}-ecs_metadata"]
+==== Event Metadata and the Elastic Common Schema (ECS)
+In addition to decoding the events, this input will add HTTP headers of the connection to each event.
```
karenzone, line 37:
```suggestion
In addition to decoding the events, this input will add HTTP headers containing connection information to each event.

```

docs/index.asciidoc
```diff
@@ -32,6 +32,28 @@ This input can also be used to receive webhook requests to integrate with other
...
+When ECS compatibility is disabled, the headers are stored in the `headers` field, which had the potential to create confusion and schema conflicts downstream.
```
karenzone, line 38:
```suggestion
When ECS compatibility is disabled, the headers are stored in the `headers` field, which has the potential to create confusion and schema conflicts downstream.
```

docs/index.asciidoc
```diff
@@ -32,6 +32,28 @@ This input can also be used to receive webhook requests to integrate with other
...
+When ECS is enabled, we can ensure a pipeline still has access to this metadata throughout the event's lifecycle without polluting the top-level namespace.
```
karenzone, line 39:
```suggestion
When ECS is enabled, we can ensure a pipeline maintains access to this metadata throughout the event's lifecycle without polluting the top-level namespace.
```

docs/index.asciidoc
```diff
@@ -115,6 +138,72 @@ and no codec for the request's content-type is found
...
+** `disabled`: unstructured connection metadata added at root level
+** `v1`: headers added under `[@metadata][http][header]`. Some are copied to structured ECS fields `http`, `url`, `user_agent` and `host`
+
+Controls this plugin's compatibility with the
+{ecs-ref}[Elastic Common Schema (ECS)].
+See <<plugins-{type}s-{plugin}-ecs_metadata>> for detailed information.
+
+Example output:
+
+ECS disabled
```
karenzone, line 155:
```suggestion
**Sample output: ECS disabled**
```

docs/index.asciidoc
```diff
@@ -115,6 +138,72 @@ and no codec for the request's content-type is found
...
+         "content_length" => "5",
+           "content_type" => "application/x-www-form-urlencoded"
+    },
+    "@timestamp" => 2021-05-28T19:27:28.609Z,
+    "host" => "127.0.0.1",
+    "message" => "hello"
+}
+-----
+
+ECS enabled
```
karenzone, line 176:
```suggestion
**Sample output: ECS enabled**
```

karenzone: Thanks for these wonderful examples.  I suggested a formatting change to give them more emphasis.

karenzone (commented):
Nice!  Left suggestions inline for your consideration.

PR #140, opened 2021-06-18
Update Netty dependency to 4.1.65

### LINKED ISSUE #139: Upgrade the version of netty
logstash-input-http currently builds with io.netty:netty-all version 4.1.49

The current version of netty at the time of writing is 4.1.63 and fixes a few security vulnerabilities. None of the issues are exploitable or directly affect logstash-input-http, but security scanners tag the findings which generates friction when having to explain the results

andsel:
Update Netty dependency to 4.1.65

Fixes #139


jsvd (approved):
LGTM please bump the patch version so we can publish this

PR #149, opened 2021-12-20
Build: do not package log4j-api dependency

kares:
The packaged log4j-api dependency is never to be used and thus no need for it to be packaged as part of the plugin.

Logstash core provides the logging dependency and should not have a second .jar with the same classes around.

---

```diff
diff --git lib/logstash-input-http_jars.rb lib/logstash-input-http_jars.rb
index 136c734..548ef9b 100644
--- lib/logstash-input-http_jars.rb
+++ lib/logstash-input-http_jars.rb
@@ -2,5 +2,4 @@
 
 require 'jar_dependencies'
 require_jar('io.netty', 'netty-all', '4.1.65.Final')
-require_jar('org.apache.logging.log4j', 'log4j-api', '2.17.0')
 require_jar('org.logstash.plugins.input.http', 'logstash-input-http', '3.4.4')

```


kaisecheng (approved):
LGTM. I tested locally and it logs normally and removes log4j-api.jar

yaauie (approved):
👍🏼

PR #150, opened 2022-03-18
Doc: Clarify description and make minor grammar fixes

karenzone:


karenzone (commented):
Tested content in local doc build for validation. Builds and renders as expected.

PR #151, opened 2022-03-21
Feat: review and deprecate ssl protocol/cipher settings

kares:
following up on #146:
- deprecating `cipher_suites` in favor of `ssl_cipher_suites` (with better validation of the supported cipher set)
- adding `ssl_supported_protocols` as a replacement for `tls_min_version` / `tls_max_version`


lib/logstash/inputs/http.rb
```diff
@@ -236,19 +241,37 @@ def validate_ssl_settings!
...
       @ssl_verify_mode_final = @ssl_verify_mode
     end
 
+    if @ssl && (original_params.key?('cipher_suites') && original_params.key?('ssl_cipher_suites'))
+      raise LogStash::ConfigurationError, "Both `ssl_cipher_suites` and (deprecated) `cipher_suites` were set. Use only `ssl_cipher_suites`."
+    elsif original_params.key?('cipher_suites')
+      @ssl_cipher_suites_final = @cipher_suites
+    else
+      @ssl_cipher_suites_final = @ssl_cipher_suites
+    end
```
andsel, line 257: I think that if `@ssl` is `false` then `@ssl_cipher_suites_final` end in:
- empty list if the user configured `cipher_suites`.
-  `SslSimpleBuilder.getDefaultCiphers` if the user in any other case. 
I think that if `ssl => false` it shouldn't set anything related to the ciphers

 I know that the method `build_ssl_params` has a guard:
```ruby
return nil unless @ssl
```
Maybe isolating the fragment with an `if @ssl`should help.

```suggestion
    if @ssl
      if original_params.key?('cipher_suites') && original_params.key?('ssl_cipher_suites')
        raise LogStash::ConfigurationError, "Both `ssl_cipher_suites` and (deprecated) `cipher_suites` were set. Use only `ssl_cipher_suites`."

      @ssl_cipher_suites_final = original_params.key?('cipher_suites') ? 
            @cipher_suites : 
            @ssl_cipher_suites
    end
```

lib/logstash/inputs/http.rb
```diff
@@ -236,19 +241,37 @@ def validate_ssl_settings!
...
+      raise LogStash::ConfigurationError, "Both `ssl_supported_protocols` and (deprecated) `tls_min_ciphers` were set. Use only `ssl_supported_protocols`."
+    elsif @ssl && (original_params.key?('tls_max_version') && original_params.key?('ssl_supported_protocols'))
+      raise LogStash::ConfigurationError, "Both `ssl_supported_protocols` and (deprecated) `tls_max_ciphers` were set. Use only `ssl_supported_protocols`."
+    else
+      if @ssl && (original_params.key?('tls_min_version') || original_params.key?('tls_max_version'))
+        @ssl_supported_protocols_final = TLS.get_supported(tls_min_version..tls_max_version).map(&:name)
+      else
+        @ssl_supported_protocols_final = @ssl_supported_protocols
+      end
+    end
```
andsel, line 269: Same observation as the one above, protect with an `if @ssl` could make it more readable.

andsel (changes requested):
Tested locally, with deprecated config settings and the plugin starts cleanly with only the warnings, it doesn't create any breakage.

I've requested changes to know your opinion on those, if you feel it's worthwhile to simplify.

lib/logstash/inputs/http.rb
kares: `ssl_cipher_suites_final` won't be used since `build_ssl_params` isn't called.
the reason why I have it this way is due existing convention above: 
```ruby
    # ...
    elsif original_params.key?("verify_mode")
      @ssl_verify_mode_final = @verify_mode
    else
      @ssl_verify_mode_final = @ssl_verify_mode
```

lib/logstash/inputs/http.rb
kares: if the attr is not set `@ssl_supported_protocols_final = @ssl_supported_protocols` is just `nil`

kares: here's a compromise refactoring: https://github.com/logstash-plugins/logstash-input-http/pull/151/commits/db159bbfe07de78546bb3f2430fd0e616a181249 ... wdyt?

andsel (changes requested):
Left just maybe a nitpick at https://github.com/logstash-plugins/logstash-input-http/pull/151#discussion_r852847059 but I think it makes the code more readable.

Everything else seems ok to me.

andsel (approved):
LGTM

PR #153, opened 2022-04-12
Codec pipeline context

yaauie:
This PR does not change the implementation, and therefore doesn't need a `CHANGELOG` entry or `VERSION`-bump.

1. In #152 I committed a line that 🤦🏼 bypassed most specs. This PR backtracks that one line and specs remain green
2. In #152 we solve the issue where execution context isn't propagated to `additional_codecs`, but the spec incorrectly validated with a _global_ setting value. Since Logstash's `ECSCompatibilitySupport` _falls through_ to the global settings, this spec wasn't validating the actual issue. I have refactored the spec to use a pipeline setting override, and proved that the solution as shipped in 3.5.1 does solve the issue.

### DIFF
--- a/spec/inputs/http_spec.rb
+++ b/spec/inputs/http_spec.rb
@@ -664,19 +664,11 @@ end
 # respected.
 if LogStash::SETTINGS.registered?('pipeline.ecs_compatibility')
 
-  def with_setting(name, value, &block)
-    setting = LogStash::SETTINGS.get_setting(name)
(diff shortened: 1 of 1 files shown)


jsvd (approved):
LGTM, good thing you spotted this!
One more reason to make all of these tests suites to no longer rely on a pipeline..there's a lot of wiring that can easily break or produce wrong results.

PR #160, opened 2022-11-11
fix: provide ssl engine with advisory peer and algorithm info

### LINKED ISSUE #159: [stub] add advisory hints to ssl engine to fast-track negotiation
This is a stub issue for tracking to be followed-up with a PR that describes the improvement more succinctly with code than I can do in a formal issue description.

yaauie:
Resolves #159


yaauie:
It turns out my expectations were not quite right, and this hint should be provided with an opt-in flag.

I will come back to it shortly.

PR #162, opened 2023-01-19
Update Netty to 4.1.87

andsel:
## Release notes
Update Netty to `4.1.87`

## What does this PR do?

With Netty's [PR 11732](https://github.com/netty/netty/pull/11732), Netty projects stops to re-package jars in `netty-all` dependency, so all the Netty deps used by the project has to be explicited listed, and this started with version `4.1.69`.
With version [`4.1.78`](https://netty.io/news/2022/06/14/4-1-78-Final.html) was shipped the [PR 12438](https://github.com/netty/netty/pull/12438/files) which in `SslContext` needs the unix socket domains, packaged in `netty-transport-native-unix-common`.

## Why is it important/What is the impact to the user?

Keeps the dependencies updated.


kaisecheng (approved):
LGTM

PR #167, opened 2023-06-29
bump netty to 4.1.94

jsvd:


jsvd:
@logstashmachine bump patch

PR #171, opened 2023-11-20
[WIP] Fix keystore and truststore support

edmocosta:


spec/fixtures/certs/generate.sh
```diff
@@ -12,11 +12,13 @@ echo "DO NOT USE THESE CERTIFICATES IN PRODUCTION" >> ./README.txt
 # certificate authority
 openssl genrsa -out root.key 4096
 openssl req -new -x509 -days 1826 -extensions ca -key root.key -out root.crt -subj "/C=LS/ST=NA/L=Http Input/O=Logstash/CN=root" -config ../openssl.cnf
+keytool -import -file root.crt -alias rootCA -keystore truststore.jks -noprompt -storepass 12345678
```
jsvd, line 15: As I was testing I noticed that creating the root CA p12 trust store using openssl like this:

`openssl pkcs12 -export -in root.crt -nokeys -out root_cert.p12 -name root -passout pass:root_cert`

produced a store that couldn't be read correctly, while using the same parameters with `keytool` like this:

`keytool -import -alias root -noprompt -file root.crt -keystore root_cert.p12 -storetype PKCS12 -storepass root_cert`

worked fine. was this also why you used keytool here?

jsvd: Seems to be related to https://github.com/openssl/openssl/pull/19025
and the fact that openssl wont, by default, set the trusted cert bag attribute, while the keytool does:

```
❯ openssl pkcs12 -in root_cert_openssl.p12 -nokeys -passin pass:root_cert | grep Bag
Bag Attributes: <No Attributes>
```

```
❯ openssl pkcs12 -in root_cert_keytool.p12 -nokeys -passin pass:root_cert | grep Bag -A 2
Bag Attributes
    friendlyName: root
    2.16.840.1.113894.746875.1.1: <Unsupported tag 6>
```

This goes deep into p12 vs jks and java interoperatiblity, I think it's ok for now that we stick to using the keytool to create the trust store.

edmocosta: Yes, I had the same issue. I'm not pretty sure why that's happening - and TBH, didn't dig into it - but it seems that when it's generated using `openssl`, the trusted certificated is not included/readable on/from the keystore. It might be something related to how manticore reads it, encryption compatibility, etc. as it also uses the `keytool` to generate their [truststore](https://github.com/cheald/manticore/blob/master/Rakefile#L54) test files.

src/main/java/org/logstash/plugins/inputs/http/util/SslSimpleBuilder.java
```diff
@@ -148,28 +204,80 @@ public boolean isClientAuthenticationRequired() {
...
+                    logger.debug("Certificate Authorities: {}", Arrays.toString(certificateAuthorities));
+                }
 
-            builder.clientAuth(verifyMode.toClientAuth())
-                    .trustManager(loadCertificateCollection(certificateAuthorities));
+                builder.trustManager(loadCertificateCollection(certificateAuthorities));
+            } else if (trustStore != null || keyStore != null) {
+                builder.trustManager(createTrustManagerFactory());
+            } else {
+                throw new IllegalStateException("Either an SSL certificate or an SSL Trust Store is required when SSL is enabled");
```
jsvd, line 231: I don't believe this should ever happen, right? if `ssl_enabled => true` then we already required the ssl_certificate or keystore to be set here:

```ruby
     unless ssl_certificate_configured? || ssl_keystore_configured?
      raise LogStash::ConfigurationError, "Either an `ssl_certificate` or `ssl_keystore_path` is required when SSL is enabled `#{ssl_config_name} => true`"
    end
```

src/main/java/org/logstash/plugins/inputs/http/util/SslSimpleBuilder.java
```diff
@@ -129,16 +162,39 @@ public SslSimpleBuilder setCipherSuites(String[] ciphersSuite) throws IllegalArg
...
+        }
+
+        this.trustStore = readKeyStore(
+                trustStorePath,
+                resolveKeyStoreType(trustStoreType, trustStorePath),
+                formatJksPassword(trustStorePassword)
+        );
+
+        if (!hasTrustStoreEntry(this.trustStore)) {
+            logger.warn("The provided Trust Store file does not contains any trusted certificate entry: {}. Please confirm this is the correct certificate and the password is correct", trustStoreFile);
```
jsvd, line 192: I wonder if here we should differentiate between our expectations from a key store being used as a trust store vs one doubling as both key store and trust store.
For a "trust store"-only p12 I'd expect that only certificates live in it, and any keys would be a mistake. on the other hand, when reusing a key store as a trust store anything goes, and should probably be discouraged in production.
I'd suggest checking how our other products do it, but in either case leaving here a comment saying that `hasTrustStoreEntry` could be made to verify either in a strict or loose modes.

src/main/java/org/logstash/plugins/inputs/http/util/SslSimpleBuilder.java
```diff
@@ -129,16 +162,39 @@ public SslSimpleBuilder setCipherSuites(String[] ciphersSuite) throws IllegalArg
...
-            throw new IllegalArgumentException("Certificate authorities are required to enable client authentication");
+    public SslSimpleBuilder setClientAuthentication(SslClientVerifyMode verifyMode) {
+        this.verifyMode = verifyMode;
+        return this;
+    }
+
+    public SslSimpleBuilder setCertificateAuthorities(String[] certificateAuthorities) {
+        if (certificateAuthorities == null || certificateAuthorities.length == 0){
+            throw new IllegalArgumentException("SSL certificate authorities is required");
         }
```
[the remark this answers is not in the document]
edmocosta: It shouldn't be possible, it's just an input's validation to ensure the builder's state is right independently of how its used externally.

src/main/java/org/logstash/plugins/inputs/http/util/SslSimpleBuilder.java
edmocosta: Exactly, it shouldn't happen, unless a bug is introduced. Same reasoning as [this](https://github.com/logstash-plugins/logstash-input-http/pull/171/files#r1406412378)

src/main/java/org/logstash/plugins/inputs/http/util/SslSimpleBuilder.java
```diff
@@ -148,28 +204,80 @@ public boolean isClientAuthenticationRequired() {
...
+        return KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+    }
+
+    private TrustManagerFactory createTrustManagerFactory() throws Exception {
+        final TrustManagerFactory tmf = getDefaultTrustManagerFactory();
+        if (this.trustStore == null) {
+            logger.info("SSL Trust Store not configured, using the provided Key Store instead.");
+
+            if (logger.isDebugEnabled() && !hasTrustStoreEntry(this.keyStore)) {
+                logger.debug("The provided SSL Key Store, used as Trust Store, has no trusted certificate entry.");
```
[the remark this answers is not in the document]
edmocosta: See: https://github.com/logstash-plugins/logstash-input-http/pull/171/files#r1406458675

src/main/java/org/logstash/plugins/inputs/http/util/SslSimpleBuilder.java
jsvd: Overall I'm concerned that we let users get to a state that won't work. if we raised an exception here, now, I don't believe there would be any breaking environments that would be working before. We would have to, however, only raise exception if the client auth is set to required.

edmocosta: >On the final part of the sentence, do we know of an invalid password situation that hits this code? from my test setting a wrong password in the trust store shows a different exception:

That message is indeed confusing. It's coming from the JDK when it tries to read the `truststore` file [here](https://github.com/logstash-plugins/logstash-input-http/pull/171/files/dd8e7d4e366ff0d9d30e4dc5792005bf5275f87b#diff-cd477b6c3d223bfe35f4076bec10e6240dbfaf88d5d714c91e90e80e93a93700R185) with an invalid password. I think I can catch this exception and re-throw it with a more meaningful message to the user. WDYT?

>Overall I'm concerned that we let users get to a state that won't work. if we raised an exception here, now, I don't believe there would be any breaking environments that would be working before. We would have to, however, only raise exception if the client auth is set to required.

Agreed, this concern also applies to other plugins, that currently have minimum to no validation regarding it. Considering it's a new functionally, I agree with you! It's better to avoid those invalid certificates from the very beginning!  I'll change it :)

jsvd: > That message is indeed confusing. It's coming from the JDK when it tries to read the truststore file [here](https://github.com/logstash-plugins/logstash-input-http/pull/171/files/dd8e7d4e366ff0d9d30e4dc5792005bf5275f87b#diff-cd477b6c3d223bfe35f4076bec10e6240dbfaf88d5d714c91e90e80e93a93700R185) with an invalid password. I think I can catch this exception and re-throw it with a more meaningful message to the user. WDYT?

just to clarify, I'm ok with the current exception that happens when the password is incorrect, but I don't think that the message we introduce in this PR is accurate by mentioning that the password may be incorrect at that point:

> logger.warn("The provided Trust Store file does not contains any trusted certificate entry: {}. Please confirm this is the correct certificate **and the password is correct**", trustStoreFile);

edmocosta: Sorry, I misunderstood your message. I've changed it to throw an exception and removed the misleading part of the message. Thanks 👍

jsvd (approved):
In the tests I've done manually everything worked correctly, both with certs directly derived from a root CA and using an intermediate CA, with jks and p12 stores.

There is more to do on ensuring good default behaviours when configuring TLS, but they're out of scope for this PR.

LGTM :shipit:

PR #172, opened 2024-04-11
Update netty version to fix GHSA-5jpm-x58v-624v

debasishbsws:
The version of netty you are using is affected with a CVE GHSA-5jpm-x58v-624v
Updating its version hoping the CVE will get fix for this.


debasishbsws:
Thanks for updating the package version

PR #176, opened 2024-08-02
docs: add clarity around SSL settings

yaauie:
- Preface text adapted from Logstash Integration Filter (which operates in client-mode and therefore presents trust before identity, whereas this server-mode presents identity before trust)
 - Adds details about the expected encodings and contents of certificate files


edmocosta (approved):
LGTM 🎉

PR #179, opened 2024-10-01
rate-limiting: propagate back-pressure from queue as HTTP 429's

yaauie:
Adds a proactive handler that rejects new requests with HTTP 429's when the queue has been blocking for more than 10 consecutive seconds, allowing back- pressure to propagate in advance of filling up the connection backlog queue.

While I'd like to take the time to break out the supplied `ExecutionObserver` into a separate library, I think it is better to roll with what we have than to hold back.


src/main/java/org/logstash/plugins/inputs/http/util/ExecutionObserver.java
```diff
@@ -0,0 +1,201 @@
...
+        boolean markComplete() {
+            isComplete = true;
+
+            // concurrency: use plain memory for reads because we can tolerate
+            // completed nodes remaining as the result of a race
+            final Execution preCompletionNext = this.getNextPlain();
+            if (preCompletionNext != null) {
+                final Execution newNext = preCompletionNext.chaseTail();
+                return (newNext != preCompletionNext) && NEXT.compareAndSet(this, preCompletionNext, newNext);
+            }
```
yaauie, line 172: review note:

We can't actually compact the completed node out, since we don't have a reference to its prev and a doubly-linked list is more complex than we need.

But we _can_ fast-forward a newly-completed node's `next` pointer to the first node in the sequence that is either not yet completed or has no next, which may be a compaction that frees up memory and reduces future computation.

When we succeed at compacting, we return true so that the caller can make decisions about performing additional compactions.

src/main/java/org/logstash/plugins/inputs/http/util/ExecutionObserver.java
```diff
@@ -0,0 +1,201 @@
...
+ * An {@code ExecutionObserver} observes possibly-concurrent execution, and provides information about the
+ * longest-running observed execution.
+ *
+ * <p>
+ *     It is concurrency-safe and non-blocking, and uses plain memory access where practical.
+ * </p>
+ */
+public class ExecutionObserver {
+    private final AtomicReference<Execution> head; // newest execution
+    private final AtomicReference<Execution> tail; // oldest execution
```
yaauie, line 19: TODO: invert to match java semantics, where things go into the tail (back) of a queue and are removed from the head (front)

yaauie:
Rebased and targeted to 3.x branch (for LS 8.x series), will need a forward-port to 4.x/main

mashhurs (commented):
Overall implementation details looks good to me (will run on my local to test)!
Some questions I would like to clarify:
- We have thread pool with `maxPendingRequests`. Is there possibility `IMessageHandler#onNewMessage` take much time (>10s) but no pending requests in thread pool? If so, involving pending request int logic would also make sense to me.
- I wonder if we need to make hard coded 10s configurable as an expert-only config. For very intensive applications 10s may even be long as my understanding.

yaauie:
> * We have thread pool with `maxPendingRequests`. Is there possibility `IMessageHandler#onNewMessage` take much time (>10s) but no pending requests in thread pool? If so, involving pending request int logic would also make sense to me.

They are two different safeguards, and do not need to be intertwined.

If the queue is full and a write to the queue is blocking, even a _single_ request made before the block is lifted should result in a 429. The prior 429 safeguard only kicked in if there was a sufficient quantity of pending requests, and will still kick in at that threshold of concurrent pending requests regardless of blockage status. This is as-desired

> * I wonder if we need to make hard coded 10s configurable as an expert-only config. For very intensive applications 10s may even be long as my understanding.

We can make it configurable _later_, if necessary. I don't see an immediate reason to make it configurable now, especially as making the option _intuitive_ for a user to configure would be very difficult. The 10s threshold is meant to make it reactive _enough_ in a fully-blocked state without causing superfluous 429's for intermittent blockages, _especially_ since most connecting clients (unfortunately) don't use http's `Expect: 100-continue` and therefore would need to retransmit data.

mashhurs (approved):
LGTM! Thank you so much for this, it is very helpful.

PR #182, opened 2024-11-13
Remove deprecated SSL settings

robbavey:
This commit marks SSL settings `obsolete` that were previously marked as `deprecated` as part

of the SSL Settings Standardization process implemented in `3.7.0` of this plugin.

Marking these settings as `obsolete` is a *breaking change* that will stop the plugin from starting, and forces users to move to the new standard settings.

This commit makes the following configuration settings obsolete:

This commit cleans up some code to handle duplicate settings, and removes tests that were put in place to support the co-existence of deprecated and non-deprecated settings, replacing them with tests that verify that obsolete settings are identified early, and information about the deprecation is related to the user.

Thanks for contributing to Logstash! If you haven't already signed our CLA, here's a handy link: 

Relates: #181


lib/logstash/inputs/http.rb
```diff
@@ -150,25 +134,15 @@ class LogStash::Inputs::Http < LogStash::Inputs::Base
...
-  config :cipher_suites, :validate => :array, :default => [], :deprecated => "Set 'ssl_cipher_suites' instead."
-
-  # The minimum TLS version allowed for the encrypted connections. The value must be one of the following:
-  # 1.0 for TLS 1.0, 1.1 for TLS 1.1, 1.2 for TLS 1.2, 1.3 for TLS 1.3
-  config :tls_min_version, :validate => :number, :default => TLS.min.version, :deprecated => "Set 'ssl_supported_protocols' instead."
-
-  # The maximum TLS version allowed for the encrypted connections. The value must be the one of the following:
-  # 1.0 for TLS 1.0, 1.1 for TLS 1.1, 1.2 for TLS 1.2, 1.3 for TLS 1.3
-  config :tls_max_version, :validate => :number, :default => TLS.max.version, :deprecated => "Set 'ssl_supported_protocols' instead."
+  # Obsolete Settings
```
donoghuc, line 137: Did we decide that doing validations on obsolete settings is potentially confusing to users. https://github.com/logstash-plugins/logstash-input-elasticsearch/pull/213#discussion_r1842794333 (with the caveat that we still use obfuscated for sensitive data)?

robbavey: Good catch - you're correct. I've updated to make the changes

donoghuc (approved):
LGTM, CI appears to be backed up. Assuming that gets through green this should be g2g :)

donoghuc:
I think test failures are due to expired cert fixtures. Regenerated in https://github.com/logstash-plugins/logstash-input-http/pull/184

donoghuc (approved):
All green after rebase! 🍏

robbavey:
Ready for docs review @karenzone

karenzone (approved):
LGTM

PR #191, opened 2025-02-03
Name netty threads properly.

mashhurs:
Inspired by https://github.com/logstash-plugins/logstash-input-beats/pull/511
Namning netty threads to make investigation easier.


donoghuc (approved):
LGTM

Issue #200, opened 2025-09-04
Triage flakey tests


donoghuc:
Looking at the build history, random cells fail very often. We should figure out the cause of the flaky tests and make them more reliable. 

**Applicable branches**:
- main
- 3.x

**Failure history**:

- https://app.travis-ci.com/github/logstash-plugins/logstash-input-http/builds

donoghuc:
The main culprit is this:

https://github.com/logstash-plugins/logstash-input-http/blob/0d44b86d1b5955ae105d9d7ad50d3cdf57f8c6c4/src/test/java/org/logstash/plugins/inputs/http/util/ExecutionObserverTest.java#L160
```
w Enable Watchlogstash-1  | <===========--> 85% EXECUTING [3s]> :test > 1 test completed> :test > Executing test org.logstash.plugins.inputs.http.util.ExecutionObservew Enable Watchlogstash-1  | ExecutionObserverTest > testThreadSafetyBruteForce() > repetition 1 of 10 FAILED
w Enable Watchlogstash-1  |     java.lang.AssertionError: 
w Enable Watchlogstash-1  |     Expected: is (a value greater than <0> and a value less than <10>)
w Enable Watchlogstash-1  |          but: a value less than <10> <10> was equal to <10>
w Enable Watchlogstash-1  |         at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
w Enable Watchlogstash-1  |         at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:6)
w Enable Watchlogstash-1  |         at org.logstash.plugins.inputs.http.util.ExecutionObserverTest.testThreadSafetyBruteForce(ExecutionObserverTest.java:197)
w Enable Watchlogstash-1  | 
w Enable Watchlogstash-1  | 
w Enable Watchlogstash-1  | 
```

PR #201, opened 2025-10-20
Fix flaky assertion bound in testThreadSafetyBruteForce

donoghuc:
Local compaction depends on completion order and timing. Under certain cases node count can reach or exceed `sqrt(concurrency)` before the final compaction is triggered. The test for intermediate state was failing fairly often in CI. This commit updates the test for intermediate state to simply be less than `concurrency`. This should satisfy the intent of the test, with the added security of the rest of the test asserting the finalized state is less than 2 nodes.

Closes https://github.com/logstash-plugins/logstash-input-http/issues/200


src/test/java/org/logstash/plugins/inputs/http/util/ExecutionObserverTest.java
```diff
@@ -194,7 +194,7 @@ void testThreadSafetyBruteForce() {
             // without queries, we should at least have some compaction
```
yaauie:
```suggestion
            // without queries, we may have some compaction so we shouldn't be too strict
```

PR #221, opened 2026-06-29
bind in register

### LINKED ISSUE #215: Bind to port during `#register`
When we fail to bind to the port, we should propagate the failure to prevent the pipeline from starting.

With this netty-based plugin, the pattern should be:
 - bind in `#register` with the server channel set to _disable_ auto-read
 - enable auto-read in `#run` once the plugin has a reference to the queue writer

Relates: elastic/logstash#18711

yaauie:
Binds to the port in `#register`, so that a failure to bind can be surfaced to the user.

Resolves: #215


lib/logstash/inputs/http.rb
```diff
@@ -190,6 +190,12 @@ def register
     require "logstash/inputs/http/message_handler"
     message_handler = MessageHandler.new(self, @codec, @codecs, @auth_token)
     @http_server = create_http_server(message_handler)
+    begin
+      logger.debug("Binding http input to port", :address => "#{@host}:#{@port}", :ssl_enabled => @ssl_enabled)
+      @http_server.bind
+    rescue java.net.BindException => bind_exception
+      fail LogStash::ConfigurationError, "could not bind to #{@host}:#{@port}; #{bind_exception.message}"
```
donoghuc, line 198:
```suggestion
      @http_server.close rescue nil
      fail LogStash::ConfigurationError, "could not bind to #{@host}:#{@port}; #{bind_exception.message}"
```
Looks like error on register wont close in https://github.com/elastic/logstash/blob/0875e0b103b8ae9c856259376b2c8232af5d37a3/logstash-core/lib/logstash/java_pipeline.rb#L251-L260 so do it here?

donoghuc (changes requested):
One small question. Otherwise this looks good.

donoghuc (approved):

Issue #132, opened 2020-11-30
Error in handling charsets different from UTF-8


andsel:
- Version: 3.3.5
- Operating System:
- Config File (if you have sensitive info, please remove it):
```
input {
	http {
		port => 9006
		codec => plain {
			charset => "CP1254"
		}
	}
}	

output {
	stdout {
		codec => json {charset => "UTF-8"}
	}
}
```
- Sample Data:
```python
import requests
API_ENDPOINT = "http://127.0.0.1:9006"
message='TÜRKÇE karakter test : ĞÜŞİÇÖışüğöç'
r = requests.post(url = API_ENDPOINT, data = bytes(message,'cp1254'))
```
- Steps to Reproduce:
 - run logstash with the pipeline
 - execute the python script
- the console output is:
```
{"message":"Tï¿½RKï¿½E karakter test : ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½ï¿½","@version":"1","@timestamp":"2020-11-30T10:38:55.338Z","headers":{"connection":"keep-alive","request_method":"POST","http_accept":"*/*","http_user_agent":"python-requests/2.21.0","content_length":"35","http_version":"HTTP/1.1","http_host":"127.0.0.1:9006","request_path":"/","accept_encoding":"gzip, deflate"},"host":"127.0.0.1"}
```

Is not a problem in the codec because I've tried with this pipeline:
```
input {
	file {
		path => "/tmp/cp1254_encoded.txt"
		mode => "read"
		sincedb_path => "/dev/null"
		file_completed_log_path => "/tmp/file_actions.log"
		file_completed_action => "log"
		codec => plain {
			charset => "CP1254"
		}
	}
}	

output {
	stdout {
		codec => json {charset => "UTF-8"}
	}
}
```

And the file attached
[cp1254_encoded.txt](https://github.com/logstash-plugins/logstash-input-http/files/5615435/cp1254_encoded.txt)

and the console out is what's expected (`TÜRKÇE karakter test : ĞÜŞİÇÖışüğöç`)

NB:
to reproduce the text file simply cut&paste the above string in a text editor and ask the editor to save it with encoding `CP1254`

GokcerBelgusen:
Hi guys, any improvement about this issue ?

GokcerBelgusen:
[...]

PR #230, opened 2026-07-28
Use content-type header charset to decode the body

### LINKED ISSUE #231: Use content-type charset to decode the body of post request
Relates #132 

**Logstash information**:

Please include the following information:

1. Logstash version (e.g. `bin/logstash --version`)
2. Logstash installation source (e.g. built from source, with a package manager: DEB/RPM, expanded from tar or zip archive, docker)
3. How is Logstash being run (e.g. as a service/service manager: systemd, upstart, etc. Via command line, docker/kubernetes)
4. How was the Logstash Plugin installed

**JVM** (e.g. `java -version`):

If the affected version of Logstash is 7.9 (or earlier), or if it is NOT using the bundled JDK or using the 'no-jdk' version in 7.10 (or higher), please provide the following information:

1. JVM version (`java -version`)
2. JVM installation source (e.g. from the Operating System's package manager, from source, etc).
3. Value of the `JAVA_HOME` environment variable if set.

**OS version** (`uname -a` if on a Unix-like system):

**Description of the problem including expected versus actual behavior**:
When an HTTP request contains the `Content-type` and that specify also the charset of the body, that charset has to be used to decode the body.

**Steps to reproduce**:

 1. Create a file with some payload in `CP1254`

    ```shell
      printf '{"msg":"TÜRKÇE karakter test : ĞÜŞİÇÖışüğöç"}' | iconv -f UTF-8 -t cp1254 > /tmp/payload.json
    ```
 2. Run logstash with HTTP input pipeline as:
    ```
    input  { http { port => 5056 } }
    output { stdout { codec => rubydebug } }
    ```
 3. Send the payload with:
    ```shell
    curl -X POST localhost:5056 -H 'Content-Type: application/json; charset=CP1254' --data-binary @/tmp/payload.json
    ```

**Provide logs (if relevant)**:
The Logstash event should include the `msg` as:
```
"msg" => "TÜRKÇE karakter test : ĞÜŞİÇÖışüğöç"
```

but instead it is:
```
"msg" => "T�RK�E karakter test : ������������",
```

andsel:
Change MessageProcessor to use the charset contained in Content-type header if available, else fallback on UTF-8.

- Relates #132
- Closes #231


src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
```diff
@@ -3,19 +3,13 @@
...
 import io.netty.channel.ChannelHandlerContext;
-import io.netty.handler.codec.http.DefaultFullHttpResponse;
-import io.netty.handler.codec.http.DefaultHttpHeaders;
-import io.netty.handler.codec.http.FullHttpRequest;
-import io.netty.handler.codec.http.FullHttpResponse;
-import io.netty.handler.codec.http.HttpHeaderNames;
-import io.netty.handler.codec.http.HttpHeaders;
-import io.netty.handler.codec.http.HttpResponse;
-import io.netty.handler.codec.http.HttpResponseStatus;
+import io.netty.handler.codec.http.*;
```
jsvd: please maintain these imports

src/main/java/org/logstash/plugins/inputs/http/MessageProcessor.java
```diff
@@ -73,7 +74,9 @@ public void run() {
 
     private FullHttpResponse processMessage() {
         final Map<String, String> formattedHeaders = formatHeaders(req.headers());
-        final String body = req.content().toString(UTF8_CHARSET);
+        Charset charset = HttpUtil.getCharset(req, UTF8_CHARSET);
+//        final String body = req.content().toString(UTF8_CHARSET);
```
donoghuc:
```suggestion
```

donoghuc (approved):
Left a suggestion to delete commented out code, but otherwise this looks correct. Thanks for adding a test.

donoghuc:
Also, will this go in the 3.x branch as well?

spec/inputs/http_spec.rb
```diff
@@ -223,6 +223,17 @@
...
+      context "when receiving an application/json request with charset ISO-8859-1 in content-type" do
+        it "should decode the body using the charset declared in the content-type header" do
+          body = '{"msg":"A Coruña"}'.encode("ISO-8859-1")
+          client.post("http://127.0.0.1:#{port}/meh.json",
+                      :headers => { "content-type" => "application/json; charset=ISO-8859-1" },
+                      :body => body).call
+          event = logstash_queue.pop
+          expect(event.get("msg")).to eq("A Coruña")
+        end
+      end
```
jsvd: it could be nice to refactor so that both charset tests are together, something like:

```ruby
    context "when receiving a content-type with a charset" do
      subject { LogStash::Inputs::Http.new("port" => port,
                                           "additional_codecs" => { "application/json" => "plain" }) }
      let(:headers) { { "content-type" => "application/json; charset=#{charset}"} }
      let(:body) { '{"msg":"A Coruña"}'.encode(charset) }

      context "UTF-8" do
        let(:charset) { "UTF-8".downcase }
        it "should decode the message accordingly" do
          client.post("http://127.0.0.1:#{port}/meh.json", :headers => headers, :body => body).call
          event = logstash_queue.pop
          expect(event.get("message")).to eq(body)
        end
      end

      context "ISO-8859-1" do
        let(:charset) { "ISO-8859-1" }
        it "should decode the message accordingly" do
          client.post("http://127.0.0.1:#{port}/meh.json", :headers => headers, :body => body).call
          puts headers.inspect
          event = logstash_queue.pop
          expect(event.get("message")).to eq(body)
        end
      end
    end
```

I haven't tested this enough, though.

jsvd (approved):
overall LGTM with Cas' suggested cleanup. I also left a nitpick that the 2 charset tests could be refactored to be under the same "context"

jsvd (approved):
LGTM

PR #233, opened 2026-08-05
Fix flaky 429 tests

kaisecheng:
## Summary

The HTTP 429 rate-limiting test ("rejects additional incoming requests with HTTP 429") is flaky across all CI stacks and branches due to race conditions in the test setup.

**Root cause:**
- `client_options` uses **string keys** (`"socket_timeout" =>`) but Manticore's constructor reads **symbol keys** (`:socket_timeout`). All timeout values silently fall back to the 10-second default. The intended 100ms socket timeout never took effect.
- `rand()` for `threads`, `max_pending_requests`, and `logstash_queue_size` produces non-deterministic values. Larger values require more concurrent blocking requests to arrive at the server within a tight window.
- `sleep 1` is a race condition. There is no guarantee that all blocking requests have reached the server and filled the executor pool + ArrayBlockingQueue within 1 second.

## Failed Test URLs

https://github.com/logstash-plugins/logstash-input-http/actions/runs/30991406140/job/92258164314
https://github.com/logstash-plugins/logstash-input-http/actions/runs/30894496793/job/91944063697

### DIFF
--- a/spec/inputs/http_spec.rb
+++ b/spec/inputs/http_spec.rb
@@ -74,46 +74,37 @@ describe LogStash::Inputs::Http do
       let(:threads) { 1 }
       let(:logstash_queue) { SizedQueue.new(logstash_queue_size) }
       let(:client_options) { {
-        "request_timeout" => 1,
-        "connect_timeout" => 3,
-        "socket_timeout" => 0.2
+        request_timeout: 5,
+        connect_timeout: 3,
+        socket_timeout: 0.5
       } }
 
       let(:config) { { "port" => port, "threads" => threads, "max_pending_requests" => max_pending_requests } }
 
       context "when sending more requests than queue slots" do
         it "rejects additional incoming requests with HTTP 429" do
-          # these will queue and return 200
-          logstash_queue_size.times.each do |i|
+          # fill the pipeline queue so the next push blocks the server worker
+          logstash_queue_size.times do
             response = client.post("http://127.0.0.1:#{port}", :body => '{}').call
             expect(response.code).to eq(200)
           end
 
-          # these will block
-          blocked_calls = (threads + max_pending_requests).times.map do
-            Thread.new do
-              begin
-                {:result => client.post("http://127.0.0.1:#{port}", :body => '{}').call}
-              rescue Manticore::SocketException, Manticore::SocketTimeout => e
-                {:exception => e}
-              end
+          # a blocked request holds its slot even after the client times out, so sending
+          # sequentially fills every worker and backlog slot before the next 429
+          blocked = 0
+          response = nil
+          deadline = Time.now + 30
+          until response && response.code == 429
+            raise "server never returned 429 (blocked=#{blocked})" if Time.now > deadline
+            begin
+              response = client.post("http://127.0.0.1:#{port}", :body => '{}').call
+            rescue Manticore::SocketTimeout
+              blocked += 1
             end
           end
 
-          # wait for blocking requests to be in flight rather than using a fixed sleep
-          blocked_calls.each { |t| Thread.pass until t.status == 'sleep' || !t.alive? }
-
-          # by now we should be rejecting with 429 since the backlog is full
-          response = client.post("http://127.0.0.1:#{port}", :body => '{}').call
           expect(response.code).to eq(429)
-
-          # ensure that our blocked connections did block
-          aggregate_failures do
-            blocked_calls.map(&:value).each do |blocked|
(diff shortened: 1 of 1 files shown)


spec/inputs/http_spec.rb
```diff
@@ -100,7 +100,8 @@
             end
           end
 
-          sleep 1 # let those requests go, but not so long that our block-detector starts emitting 429's
+          # wait for blocking requests to be in flight rather than using a fixed sleep
+          blocked_calls.each { |t| Thread.pass until t.status == 'sleep' || !t.alive? }
```
donoghuc: I was skeptical that the thread would ever be in a `sleep` state here and dug in a bit. I could not find a thread ever reporting that state. I think the test passed by accident because the server keeps blocked slots occupied after the client disconnects. Instead i think we should track requests one at a time until a 429 comes back, counting the ones that time out. Since each timed-out request provably holds a slot and sends are sequential, exactly threads + max_pending_requests requests block before the 429. 


```diff
diff --git a/spec/inputs/http_spec.rb b/spec/inputs/http_spec.rb
index 760ee38..15347fd 100644
--- a/spec/inputs/http_spec.rb
+++ b/spec/inputs/http_spec.rb
@@ -83,37 +83,28 @@ describe LogStash::Inputs::Http do
 
       context "when sending more requests than queue slots" do
         it "rejects additional incoming requests with HTTP 429" do
-          # these will queue and return 200
-          logstash_queue_size.times.each do |i|
+          # fill the pipeline queue so the next push blocks the server worker
+          logstash_queue_size.times do
             response = client.post("http://127.0.0.1:#{port}", :body => '{}').call
             expect(response.code).to eq(200)
           end
 
-          # these will block
-          blocked_calls = (threads + max_pending_requests).times.map do
-            Thread.new do
-              begin
-                {:result => client.post("http://127.0.0.1:#{port}", :body => '{}').call}
-              rescue Manticore::SocketException, Manticore::SocketTimeout => e
-                {:exception => e}
-              end
+          # a blocked request holds its slot even after the client times out, so sending
+          # sequentially fills every worker and backlog slot before the next 429
+          blocked = 0
+          response = nil
+          deadline = Time.now + 30
+          until response && response.code == 429
+            raise "server never returned 429 (blocked=#{blocked})" if Time.now > deadline
+            begin
+              response = client.post("http://127.0.0.1:#{port}", :body => '{}').call
+            rescue Manticore::SocketTimeout
+              blocked += 1
             end
           end
 
-          # wait for blocking requests to be in flight rather than using a fixed sleep
-          blocked_calls.each { |t| Thread.pass until t.status == 'sleep' || !t.alive? }
-
-          # by now we should be rejecting with 429 since the backlog is full
-          response = client.post("http://127.0.0.1:#{port}", :body => '{}').call
           expect(response.code).to eq(429)
-
-          # ensure that our blocked connections did block
-          aggregate_failures do
-            blocked_calls.map(&:value).each do |blocked|
-              expect(blocked[:result]).to be_nil
-              expect(blocked[:exception]).to be_a_kind_of Manticore::SocketTimeout
-            end
-          end
+          expect(blocked).to eq(threads + max_pending_requests)
         end
       end
     end
```

donoghuc (changes requested):

spec/inputs/http_spec.rb
```diff
@@ -69,50 +69,42 @@
...
+      let(:max_pending_requests) { 1 }
+      let(:threads) { 1 }
       let(:logstash_queue) { SizedQueue.new(logstash_queue_size) }
       let(:client_options) { {
-        "request_timeout" => 0.1,
-        "connect_timeout" => 3,
-        "socket_timeout" => 0.1
+        request_timeout: 5,
+        connect_timeout: 3,
+        socket_timeout: 0.5
```
kaisecheng, line 79: Manticore uses symbol key `options.fetch(:socket_timeout)`
The original using string key "socket_timeout" is wrong

donoghuc (approved):

spec/inputs/http_spec.rb
kaisecheng: The test relies on `t.alive?` to exit the loop. Agree that "sleep" state is not useful because Manticore's blocking I/O keeps the JRuby thread in "run" state. Server keeps blocked slots occupied because client side's Socket Timeout is not something the server would care. So, I think the original slot retention setup is deterministic, not totally nonsense. 

I have updated the test to your sequential call version. The timeout setting was never effective because the wrong string key was passed.

